PR #2, opened 2015-09-22
subChannel should be able to set peer preference

ShanniLi:
r @jcorbin @Raynos @kriskowal 
cc @anson627

From https://github.com/uber/tchannel/pull/1348

### DIFF
--- a/sub_peers.js
+++ b/sub_peers.js
@@ -62,6 +62,7 @@ TChannelSubPeers.prototype.add = function add(hostPort, options) {
     }
 
     peer = topChannel.peers.add(hostPort, options);
+    peer.setPreferConnectionDirection(self.preferConnectionDirection);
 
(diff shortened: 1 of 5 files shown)


sub_peers.js
```diff
@@ -62,6 +62,7 @@ TChannelSubPeers.prototype.add = function add(hostPort, options) {
     }
 
     peer = topChannel.peers.add(hostPort, options);
+    peer.setPreferConnectionDirection(self.preferConnectionDirection);
```
Raynos: We are mutating the preference here. Which means any co-tenant peers across sub channels better have the same preference on all sub channels.

This is last write wins semantics. I would prefer a more immutable first write is final approach.

Raynos:
Looks good. Would prefer a different semantics for cotenant peers though.

PR #5, opened 2015-09-22
Indirect access to cn header thru req.callerName

jcorbin:
To future proof against v3 and lazy reqs.

r @kriskowal @Raynos


kriskowal:
:100: Much better going forward

Raynos:
DO IT.

PR #6, opened 2015-09-22
Unadvertise should drain before close

anson627:
This is a follow-up for:
https://github.com/uber/hyperbahn/pull/3

Earlier the client changes the state to unadvertised once unadvertise call is made, without checking if the callback comes back successfully. This is because hyperbahn closes the peer and causes a connection reset error. That no longer happens because of graceful draining.

r: @Raynos @ShanniLi @jcorbin

### DIFF
--- a/hyperbahn-client.js
+++ b/hyperbahn-client.js
@@ -380,8 +380,6 @@ function unadvertise(opts) {
     timers.clearTimeout(self._advertisementTimer);
     timers.clearTimeout(self.advertisementTimeoutTimer);
     self.latestAdvertisementResult = null;
-    self.state = States.UNADVERTISED;
-    self.emit('unadvertised');
     function unadvertiseInternalCb(error, result) {
         if (error) {
             self.logger.warn('HyperbahnClient: unadvertisement failure', {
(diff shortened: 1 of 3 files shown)


hyperbahn-client.js
```diff
@@ -264,7 +264,6 @@ function sendRequest(opts, endpoint, cb) {
         timeout: (opts && opts.timeout) || self.defaultTimeout,
         hasNoParent: true,
         trace: false,
-        retryLimit: 1,
```
Raynos: Keep the retryLimit: 1 please. We have manual retries that are custom here.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "tchannel",
   "description": "network multiplexing and framing protocol for RPC or parser drag racing",
   "author": "mranney@uber.com",
-  "version": "3.0.3",
+  "version": "3.0.4",
```
Raynos: Please use `bash publish.sh patch` for version bumping

jcorbin:
This broke hyperban tests; restoring integration tests under travis in #13.

jcorbin:
nvm, it wasn't clear that the breakage was known; un-revering pending https://github.com/uber/hyperbahn/pull/3

PR #9, opened 2015-09-23
Add a draining "signal" frame

jcorbin:
When a connection starts draining, send a null-id'd declined error frame.

Note the message prefix is going to become a contract, since really this would be better off as a new frame type entirely; subsequent PR will branch essentially on "if the error has null id, and is declined, and the message starts with "draining:".

r @Raynos 
cc @blampe @prashantv @junchaowu

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -663,4 +663,24 @@ function sendLazyErrorFrame(reqFrame, codeString, message) {
         codeString, message);
 };
 
+TChannelConnection.prototype._drain =
+function _drain(reason, exempt) {
+    var self = this;
+
+    TChannelConnectionBase.prototype._drain.call(self, reason, exempt);
+
+    if (self.remoteName) {
+        sendDrainingFrame();
+    } else {
+        self.identifiedEvent.on(sendDrainingFrame);
+    }
+
+    function sendDrainingFrame() {
+        self.handler.sendErrorFrame(
+            v2.Frame.NullId, null,
+            'Declined',
+            'draining: ' + self.drainReason);
+    }
+};
+
 module.exports = TChannelConnection;
--- a/connection.js
+++ b/connection.js
@@ -280,10 +280,22 @@ TChannelConnection.prototype.onErrorFrame = function onErrorFrame(errFrame) {
         }));
         return;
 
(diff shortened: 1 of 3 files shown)


connection.js
```diff
@@ -614,4 +623,24 @@ function sendLazyErrorFrame(reqFrame, codeString, message) {
...
+    if (self.remoteName) {
+        sendDrainingFrame();
+    } else {
+        self.identifiedEvent.on(sendDrainingFrame);
+    }
+
+    function sendDrainingFrame() {
+        self.handler.sendErrorFrame(
+            v2.Frame.NullId, null,
+            'Declined',
```
Raynos: Declined or NetworkError ?

jcorbin: Declined, this doesn't indicate any particular error or problem.  Think of it as a "promise to decline and eventually close", really we want a new error code for this, but are past the point of adding a new one to v2 as far as I see it.

connection.js
```diff
@@ -263,9 +263,18 @@ TChannelConnection.prototype.onWriteError = function onWriteError(err) {
...
-        self.resetAll(err);
+    // TODO: why is err optional?
+    if (!err) {
+        return;
+    }
+
+    if (err.isErrorFrame &&
+        err.codeName === 'Declined' &&
+        err.message.indexOf('draining:') === 0) {
+        return;
```
anson627: I thought if a connection is draining, we want to take it out of connection pool, so it won't be re-used for new request.

jcorbin:
Indeed @anson627, updated so that now happens

connection.js
```diff
@@ -263,9 +263,24 @@ TChannelConnection.prototype.onWriteError = function onWriteError(err) {
 TChannelConnection.prototype.onHandlerError = function onHandlerError(err) {
     var self = this;
 
-    if (err) {
-        self.resetAll(err);
+    // TODO: why is err optional?
+    if (!err) {
+        return;
```
kriskowal: Cursory audit of the code shows that this method is used only to listen on error, and the error event is always emitted with an error. :dwarf: and my :axe:

jcorbin:
Updated.

Raynos:
This feature is good.

PR #10, opened 2015-09-23
Hyperbahn client should retry using different peers

anson627:
Today hyperbahn client advertise/unadvertise doesn't tchannel request to retry underneath (which keep tracking of retried peers and avoid retry peers that already failed). The retry at hyperbahn client level can potentially keep retrying the same bad peer and never succeed. 

@ShanniLi @Raynos @jcorbin

### DIFF
--- a/hyperbahn-client.js
+++ b/hyperbahn-client.js
@@ -264,7 +264,6 @@ function sendRequest(opts, endpoint, cb) {
         timeout: (opts && opts.timeout) || self.defaultTimeout,
         hasNoParent: true,
         trace: false,
-        retryLimit: 1,
         headers: {
             cn: self.callerName
         }


Raynos:
:-1: 

let's not turn retries on. Who cares about duplicate peers when there are 1000 peers.

anson627:
it's not about duplicate peers, every time hyperbahn client calls advertise, it creates a new tchannel request, which selects the same peer. If it fails, peer score won't change. hyperbahn client waits for some time and call advertise again, a new tchannel request is created, but the same peer is selected. And of course it keeps failing and will never succeed.

anson627:
the retry inside tchannel request takes care of that, by keep tracking of peers (hostPorts) already tried and failed:
https://github.com/uber/tchannel-node/blob/e36a95312f2e5529a8648e0352a6ec656414d528/sub_peers.js#L131

Raynos:
Write a failing test !

anson627:
this is the failing test:
https://travis-ci.org/uber/hyperbahn/builds/81878718

Raynos:
This is still the wrong solution.

Joshua's drain improvements might fix the test :)

Raynos:
Ok I'll buy doing 1 retry. i think 5 is too many.

ShanniLi:
Other solutions will need to keep tracking various error frames for peers.

anson627:
Yes, Josh's draining improvement will prevent getting stuck with draining peer and fix the test.

And this PR will only fix problem for advertise request.

To avoid getting stuck with a bad peer (not draining or closing), we need a better (error based) peer score update or select mechanism.

Raynos:
@kriskowal 

> To avoid getting stuck with a bad peer, we need a better peer score mechanism

Why was it safe to rm -rf Peer.health again ?

anson627:
Will fix the retryLimit, so close this.

PR #11, opened 2015-09-23
Make drain exemption invariant per channel

jcorbin:
So that we can re-use the existing draining/drainReason touch points for remote-triggered draining (ala #9).

r @Raynos

### DIFF
--- a/channel.js
+++ b/channel.js
@@ -332,7 +332,7 @@ TChannel.prototype.setLazyHandling = function setLazyHandling(enabled) {
     }
 };
 
-TChannel.prototype.drain = function drain(reason, exempt, callback) {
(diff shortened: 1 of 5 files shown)


Raynos:
not ok 1978 service:b res[0]: expected local reset error
⨯ sending request 12      operator: equal
      expected: 'tchannel.local.reset'
      actual:   'tchannel.request.drained'
      at: checkBRes (/home/travis/build/uber/tchannel-node/test/chan_drain.js:524:24)
    ...

Breaks tests.

Raynos:
Please `self.drainExempt = ...;` in constructor

Raynos:
do it.

PR #12, opened 2015-09-23
confuse TAP less

Raynos:
I removed a bunch of `not ok` from the test blocks
and the asserts.

This makes ctrl+f "not ok" in travis 200% more effective

r: @jcorbin @kriskowal @shannili


jcorbin:
less confuse, much win

PR #16, opened 2015-09-24
Fix retryLimit off by one

anson627:
retryLimit, looking at the name, means max number of additional requests it should send. Current implementation takes initial request as first retry, and therefore is off by one.

After this fix, the behavior is:
0 no-retry (default)
1 retry once
2 retry twice
...

Fix hyperbahn as well:
https://github.com/uber/hyperbahn/pull/9

r: @Raynos @ShanniLi @jcorbin


ShanniLi:
lgtm

Raynos:
lgtm.

Raynos:
Technically breaking but whatever.

PR #17, opened 2015-09-24
Eliminate a per req/res closure allocation in v2/handler

jcorbin:
r @Raynos @rf


Raynos:
yay. lgtm.

rf:
:+1: :+1: :+1:

PR #18, opened 2015-09-24
Resend unadvertise should drain

anson627:
Resend the reverted change and see if tests are passing.

@jcorbin

### DIFF
--- a/hyperbahn-client.js
+++ b/hyperbahn-client.js
@@ -380,8 +380,6 @@ function unadvertise(opts) {
     timers.clearTimeout(self._advertisementTimer);
     timers.clearTimeout(self.advertisementTimeoutTimer);
     self.latestAdvertisementResult = null;
-    self.state = States.UNADVERTISED;
-    self.emit('unadvertised');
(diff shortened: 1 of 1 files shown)


anson627:
tchannel tests failed due to flappy drain test. It seems to be not related.
hyperbahn tests are passing:
The command "npm run hyperbahn-link-test" exited with 0.

Raynos:
Please do not merge any changes related to HyperbahnClient without npm linking tchannel into hyperbahn and running the hyperbahn-client tests in hyperbahn.

Raynos:
I've reverted this change.

anson627:
@Raynos the travis CI does a `npm run hyperbahn-link-test` to get latest hyperbahn code, link the tchannel, and run the hyperbahn test. And I saw it passed before check-in the change.

anson627:
@Raynos I was not able to reproduce the test failure locally on my machine, I guess the problem is due to that we only allow advertise request to retry one more time, but there might be two bad peers due to draining. Anyway, I will hold on and wait for Josh's real fix.

PR #22, opened 2015-09-24
WIP: add TChannelHTTP#runHTTPHandler

jcorbin:
TODO:
- [ ] probably don't use (uber-)hammock, just inline mock http req/res classes
- [ ] tests
- [ ] examples

To address @MadanThangavelu's current needs in an internal service, and based on the request proxy pattern already present in ringpop ala.

r @Raynos 
cc @MadanThangavelu @Matt-Esch @jwolski


Raynos:
Thanks for the cleanup :D

PR #23, opened 2015-09-24
IDEA: always try to pop

jcorbin:
Not sure if this is good idea or not, @Raynos ?

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -362,9 +362,10 @@ function onCallErrorFrame(errFrame) {
         if (req.res) {
             req.res.errorEvent.emit(req.res, err);
         } else {
-            // Only popOutReq if there is no call response object yet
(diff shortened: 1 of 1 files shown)


connection.js
```diff
@@ -362,10 +362,11 @@ function onCallErrorFrame(errFrame) {
         if (req.res) {
             req.res.errorEvent.emit(req.res, err);
         } else {
-            // Only popOutReq if there is no call response object yet
-            req = self.ops.popOutReq(id, err);
             req.emitError(err);
         }
+        if (self.ops.getOutReq(id)) {
```
Raynos: bad idea. Breaks streaming.

Raynos: wait what :(

Raynos:
No idea. ENEEDMORETESTS

PR #3, opened 2015-09-22
RelayHandler: add lazy support

jcorbin:
Passing a lazy-ported copy of tests/relay with only one logging assertion elided.

r @kriskowal @Raynos

PR #27, opened 2015-09-25
v2/error_response: lazy error reading

jcorbin:
Towards #3 

r @kriskowal


kriskowal:
♗ :droplet: :droplet: :droplet:

PR #28, opened 2015-09-25
Fix connection leak during peer close

anson627:
When a peer has more than one connections, peer close always leaks one.

r: @Raynos @jcorbin @ShanniLi

### DIFF
--- a/peer.js
+++ b/peer.js
@@ -164,9 +164,9 @@ TChannelPeer.prototype.close = function close(callback) {
 
     var counter = self.connections.length;
     if (counter) {
-        self.connections.forEach(function eachConn(conn) {
-            conn.close(onClose);
-        });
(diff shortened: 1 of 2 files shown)


ShanniLi:
:+1:

anson627:
to help understand a little bit, we are basically deleting while iterating:

close is emitted:
https://github.com/uber/tchannel-node/blob/d588dbe51c64f6a666fd91caca5c7702488eefc7/connection.js#L575

close causes remove:
https://github.com/uber/tchannel-node/blob/d588dbe51c64f6a666fd91caca5c7702488eefc7/peer.js#L345

remove happens:
https://github.com/uber/tchannel-node/blob/d588dbe51c64f6a666fd91caca5c7702488eefc7/peer.js#L380

jcorbin:
lgtm

PR #29, opened 2015-09-28
Update as/thrift example

kriskowal:
This is what it takes to set up a service and client and do as/thrift now. Note that the peer has to be identified before a request. Thrift is more strict now. The Thrift method name would never have worked.

Fixes https://github.com/uber/tchannel/issues/1199

### DIFF
--- a/docs/as-thrift.md
+++ b/docs/as-thrift.md
@@ -9,11 +9,11 @@ to handle encoding and decoding for you
 
 ## Thrift example
 
-`thrift/service.thrift`:
+`thrift/echo.thrift`:
 
 ```thrift
 struct EchoResult {
-    1: string value
+    1: required string value
 }
 
 service Echo  {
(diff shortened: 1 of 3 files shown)


docs/as-thrift.md
```diff
@@ -54,25 +52,36 @@ function echo(context, req, head, body, callback) {
 }
 
 server.listen(4040, '127.0.0.1', onListening);
-
 function onListening() {
+    var peer = echoChannel.peers.add('127.0.0.1:4040');
+    peer.connect();
+    peer.waitForIdentified(onIdentified);
```
Raynos: Let's not do any waitForIdentified().

No need to `peer.connect()` use `channel.waitForIdentified({ host: host }, onIdentified)`

docs/as-thrift.md
```diff
@@ -31,20 +31,18 @@ var TChannel = require('tchannel');
 var fs = require('fs');
 var path = require('path');
 
-var server = TChannel({
-    serviceName: 'server'
-});
+var server = TChannel({serviceName: 'server'});
+
 var client = TChannel();
+var echoChannel = client.makeSubChannel({serviceName: 'echo'});
```
Raynos: Use `peers: [...]` or `peers.add()`

docs/as-thrift.md
Raynos: as long as there is a peer object in the peers collection you can just `channel.request()`

kriskowal: Unfortunately, the example does not work unless you identify first. TCurl also identifies before making requests.

kriskowal: Assertion thrown here https://github.com/uber/tchannel-node/blob/master/connection_base.js#L121

kriskowal:
Figured it out. The request host option is not the way to go™. The peer option is dubious but irrelevant for this case. May follow-up to see what we can do to make the host and peer options work without pre-identification—there would need to be an assertion that the presumed address is the actual address upon identification.

In any case, updated and ready for another review @Raynos

docs/as-thrift.md
Raynos: call request on the channel. it will identify for you.

Raynos:
lgtm.

Raynos:
Needs to pass tests.

Issue #30, opened 2015-09-29
.listen() is "allowed" on subChannels, but breaks things far far away


benfleis:
If .listen() isn't allowed on subChannels, then I suggest that we enforce this policy. Currently it half-works, and causes failures down the road; if I hadn't already gotten help solving it the first time, I would have gone crazy the second time I was upgrading something to new tchannel.

Fail fast would ideal here.

Raynos:
:+1: for fail fast.

benfleis:
I tried the obvious diff, and it breaks tests w/ EADDRINUSE. I haven't had time to track it down, but it's pretty counterintuitive to me. Maybe @jcorbin or @kriskowal will see what I'm missing?

  https://github.com/uber/tchannel-node/compare/wip-subchannel-no-listen

Failure output is in benchmarks:
```
Loading config: /Users/benfleis/src/tchannel/node/.istanbul.yml
running /Users/benfleis/src/tchannel/node/benchmarks/bench_server[31807]

/Users/benfleis/src/tchannel/node/lib/event_emitter.js:140
    throw err;
          ^
TchannelServerListenFailedError: tchannel: listen EADDRINUSE
    at createError (/Users/benfleis/src/tchannel/node/node_modules/error/typed.js:31:22)
    at Object.createError [as TChannelListenError] (/Users/benfleis/src/tchannel/node/node_modules/error/wrapped.js:88:19)
    at TChannel.onServerSocketError (/Users/benfleis/src/tchannel/node/channel.js:498:22)
    at Server.onServerSocketError (/Users/benfleis/src/tchannel/node/channel.js:385:14)
    at Server.emit (events.js:95:17)
    at net.js:1047:12
    at process._tickCallback (node.js:442:13)
    at Function.Module.runMain (module.js:499:11)
    at startup (node.js:119:16)
    at node.js:929:3
child exited /Users/benfleis/src/tchannel/node/benchmarks/bench_server[31807] code 8 signal null
running /Users/benfleis/src/tchannel/node/benchmarks/multi_bench[31808]
1:          PING, 10000/0 min/max/avg/p95:  135/1328/ 707.99/1156.00   1382ms total,  7235.89 ops/sec

/Users/benfleis/src/tchannel/node/lib/event_emitter.js:140
    throw err;
          ^
TchannelServerListenFailedError: tchannel: listen EADDRINUSE
    at createError (/Users/benfleis/src/tchannel/node/node_modules/error/typed.js:31:22)
[...]

PR #31, opened 2015-10-01
Assert against chan.listen() on sub-channels

benfleis:
causes unexpected test failure, needs tweaks.

test failure here:

  https://github.com/uber/tchannel-node/issues/30#issuecomment-144597268

### DIFF
--- a/channel.js
+++ b/channel.js
@@ -561,6 +561,7 @@ TChannel.prototype.listen = function listen(port, host, callback) {
     //   available ephemeral port
     // - 127.0.0.1 is a valid host, primarily for testing
     var self = this;
+    assert(!self.topChannel, 'TChannel must listen on top channel');
     assert(!self.listened, 'TChannel can only listen once');
     assert(typeof host === 'string', 'TChannel requires host argument');
     assert(typeof port === 'number', 'TChannel must listen with numeric port');


channel.js
```diff
@@ -559,6 +559,7 @@ TChannel.prototype.listen = function listen(port, host, callback) {
     assert(typeof host === 'string', 'TChannel requires host argument');
     assert(typeof port === 'number', 'TChannel must listen with numeric port');
     assert(host !== '0.0.0.0', 'TChannel must listen with externally visible host');
+    assert(!self.topChannel, 'TChannel must listen on top channel');
```
jcorbin: This is the opposite assertion of what you want:
- chan.topChannel is not a bool
- chan.topChannel is a reference to the top channel, if chan is a sub channel

jcorbin:
Rebased and corrected your check, but yes this is a more useful form of "fail early" that we should have.

channel.js
jcorbin: Oops, I was exactly wrong... you had it right the first time.

jcorbin:
I'll take this over and complete it from here; what you missed on the test failure was that you made the benchmark listen correctly on the root channel, but then the quit/close was still happening on the sub channel.

Diving through this code path highlight other room for hardening that I'll pull out first, and an open question about do we want a similar assertion for channel close.

cc @Raynos @kriskowal

Raynos:
lgtm.

Raynos:
:+1: for `_close()` and assertions in `close()`

jcorbin:
TODO -> TODONE by splitting close and _close, will land once tests pass

benfleis:
awesome, glad to see this picked up :)

jcorbin:
TODO: this breaks hyperbahn integration tests, we need to update hyperbahn

jcorbin:
Note: to dev iterate on this:
```
$ ./hyperbahn_link_test.sh ../hyperbahn # assuming that hyperbahn is checked out at ../hyperbahn
...
assert.js:92
  throw new assert.AssertionError({
  ^
AssertionError: cannot close sub-TChannel
    at TChannel.close (/Users/joshua/Uber/tchannel-node/channel.js:852:5)
    at ServiceDispatchHandler.purgeServices (/Users/joshua/Uber/hyperbahn/service-proxy.js:9:23057)
    at purgeServices [as _onTimeout] (/Users/joshua/Uber/hyperbahn/service-proxy.js:9:23637)
    at Timer.listOnTimeout [as ontimeout] (timers.js:112:15)
...
```

Raynos:
Hmm; we close sub channels when a serviceName is gone. 

We should split this PR into two PRs; one making listen() invalid on sub channel and the other making close() invalid.

PR #32, opened 2015-10-02
TChannelThrift UX improvements

Raynos:
In passing when trying to upgrade tchannel-gen to
tchannel@3 I want to make some user interface improvements

There are a bunch of changes in this PR and some are more
"debatable" then others.

I will break anything contraversial into a seperate PR.

r: @jcorbin @kriskowal

### DIFF
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -130,6 +130,15 @@ TChannelAsThrift.prototype.request = function request(reqOptions) {
     return req;
 };
 
+TChannelAsThrift.prototype.waitForIdentified =
+function waitForIdentified(options, cb) {
+    var self = this;
+
+    assert(self.channel, 'channel is required for waitForIdentified()');
+
+    return self.channel.waitForIdentified(options, cb);
+};
+
 TChannelAsThrift.prototype.register =
 function register(channel, name, opts, handle, spec) {
     var self = this;
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -143,6 +143,16 @@ TChannelAsThrift.prototype.register =
 function register(channel, name, opts, handle, spec) {
     var self = this;
 
+    // support register(endpoint, opts, handle)
+    if (typeof channel === 'string') {
+        assert(self.channel, 'channel is required for thrift.register()');
+        spec = handle;
+        handle = opts;
+        opts = name;
+        name = channel;
+        channel = self.channel;
+    }
+
     if (!self.logger) {
         self.logger = channel.logger;
     }
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -146,6 +146,8 @@ function register(channel, name, opts, handle, spec) {
     // support register(endpoint, opts, handle)
     if (typeof channel === 'string') {
         assert(self.channel, 'channel is required for thrift.register()');
+        assert(spec === undefined, 'must have only 4 arguments');
+
         spec = handle;
         handle = opts;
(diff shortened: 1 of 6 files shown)


hyperbahn-client.js
```diff
@@ -175,6 +176,26 @@ HyperbahnClient.prototype.setReportTracing = function setReportTracing(bool) {
     self.reportTracing = bool;
 };
 
+HyperbahnClient.prototype.getThriftCodecSync =
```
kriskowal: Let’s drop Codec. Unnecessary jargon from nowhere.

Raynos: `getThriftSync()` ?

kriskowal:
Looks good to me. Could be getThriftSync or asThriftSync I thynk.

as/thrift.js
```diff
@@ -66,17 +66,13 @@ function TChannelAsThrift(opts) {
         'channel must be provided with isHealthy');
 
     if (self.isHealthy) {
-        fs.readFile(path.join(__dirname, 'meta.thrift'), 'utf8', registerHealthCheck);
+        var thriftSource = fs.readFileSync(
```
jcorbin: sync io in constructor is sad, can we instead inject the content thru opts, and then provide a `TChannelThift.somethingSync` convenience?

as/thrift.js
```diff
@@ -142,10 +138,29 @@ TChannelAsThrift.prototype.request = function request(reqOptions) {
...
+    return self.channel.waitForIdentified(options, cb);
+};
+
 TChannelAsThrift.prototype.register =
 function register(channel, name, opts, handle, spec) {
     var self = this;
 
+    // support register(endpoint, opts, handle)
+    if (typeof channel === 'string') {
+        assert(self.channel, 'channel is required for thrift.register()');
```
jcorbin: `assert(spec === undefined)` ?

as/thrift.js
Raynos: Agreed it's bad. I wanted to fix the race condition. I can also fix race condition with `setTimeout` in tests.

The same race condition applies to anyone wanting to test their health check. Not sure if there is a good fix without breaking back compat.

hyperbahn-client.js
jcorbin: :+1: codec is 🙌weird🙌

as/thrift.js
Raynos: good idea.

malandrew:
`register()` in as-thrift seems a tad overloaded now. time to change the interface to an object with keys instead of an arity of 5? I know this would result in a major version bump, but it's something work considering.

Raynos:
We've previously committed to arity of 5 as being acceptable.

hyperbahn-client.js
malandrew: Agree on codec being unnecessary jargon. `getThriftSync()`, `getThriftFile()`, `readThriftFile()`, etc. are all better. Also, get for me implies some sort of network fetching. For local disk i/o, `read` is preferable to `get` unless we plan to allow `options.thriftFile` to be something other than a local filepath.

malandrew:
LGTM, however i think that the arity of 5 on register() only makes sense if the function signature is always the same with later args being optional. the fact that the first arg can be optional, makes me feel like we should use an object instead of arity of 5 and type sniffing. But again, that's a separate discussion.

PR #33, opened 2015-10-02
Stabilize unadvertise test part one

anson627:
This is tchannel part of change to stabilize unadvertise test.
1. unadvertise state change and event emit should wait for callback to succeed
2. removeConnectionEvent should emit along with conn, so the event listener can do sth

This change need be in before hyperbahn change:
https://github.com/uber/hyperbahn/pull/23

Multiple test runs passed
```bash
for i in {1..50}; do node unadvertise.js; done
```

r: @Raynos @ShanniLi


ShanniLi:
lgtm

PR #34, opened 2015-10-02
Add errors.isFatal

jcorbin:
The idea for this started with hyperbahn relay ad.  The intention is that this function can be used by an application to say "is this error exceptional?" to guide such things as error logging (vs downgrading to a warn, info, or just silencing).

r @kriskowal @Raynos

### DIFF
--- a/errors.js
+++ b/errors.js
@@ -826,6 +826,29 @@ module.exports.toHTTPCode = function toHTTPCode(codeName) {
     }
 };
 
+module.exports.isFatal = function isFatal(err, codeName) {
+    if (!codeName) {
+        codeName = module.exports.classify(err);
+    }
+    switch (codeName) {
+        case 'Busy':
+        case 'Cancelled':
+        case 'Declined':
+        case 'NetworkError':
+        case 'Timeout':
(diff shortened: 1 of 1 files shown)


errors.js
```diff
@@ -823,6 +823,26 @@ module.exports.toHTTPCode = function toHTTPCode(codeName) {
...
+module.exports.isFatal = function isFatal(err, codeName) {
+    if (!codeName) {
+        codeName = module.exports.classify(err);
+    }
+    switch (codeName) {
+        case 'Busy':
+        case 'Cancelled':
+        case 'Declined':
+        case 'NetworkError':
+        case 'Timeout':
```
kriskowal: Unhealthy belongs up here, though we do not yet use it.

kriskowal:
:ok: pending change. Would be more critical knowing the context in which it’s intended to be used, but I expect you’re deciding whether to shut down a connection.

Raynos:
The first use case is

```js
if (errors.isFatal(err)) {
  logger.error('pagerduty');
} else {
  logger.warn('maybe bug?');
}
```

PR #35, opened 2015-10-02
Add a method to AsThrift to load Meta thrift synchronously

Raynos:
This refactors a bunch of code in AsThrift and adds a sync
helper that will deterministically load your meta thrift
file synchronously rather then asynchronously.

This fixes flappy BadRequests in the tests because of 
file IO & TCP IO race conditions.

r: @jcorbin @kriskowal


jcorbin:
:+1:

PR #37, opened 2015-10-05
Fix bugs in lazy relaying

jcorbin:
With tests to demonstrate the stack traces reported in #29.

r @kriskowal @Raynos

### DIFF
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -393,6 +393,13 @@ function _extendLogInfo(info) {
     return info;
 };
 
+LazyRelayOutReq.prototype.emitError =
+function emitError(err) {
+    var self = this;
+
+    self.inreq.onError(err);
+};
+
 LazyRelayOutReq.prototype.logError =
 function relayRequestlogError(err, codeName) {
     var self = this;
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -141,6 +141,8 @@ function LazyRelayInReq(conn, reqFrame) {
     self.timeout = 0;
     self.timedOut = false;
     self.alive = true;
+    self.operations = null;
+    self.timeHeapHandle = null;
 
     self.boundExtendLogInfo = extendLogInfo;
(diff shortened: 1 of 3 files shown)


jcorbin:
The only question I really have is do we want to keep the dead warning logs as is for now, or work to silence them?  See travis output or run test locally to see more, but basically it's things like:
```
# relay request times out
TCHANNEL WARN: dropping error from dead relay request ~ { error:
   { [TchannelRequestTimeoutError: request timed out after 96ms (limit was 93ms)]
     type: 'tchannel.request.timeout',
...
ok 17 expected timeout error
ok 18 expected no response
TCHANNEL WARN: dropping error from dead relay request ~ { error:
   { [TchannelLocalResetError: tchannel: tchannel: Connection was manually closed.]
     type: 'tchannel.local.reset',
...
# relay request declines on no peer
TCHANNEL WARN: no relay peer available ~ { requestType: 'tchannel.lazy.incoming-request',
  inRemoteAddr: '127.0.0.1:56144',
  inRequestId: 2,
  serviceName: 'dest' }
ok 19 expected declined error
ok 20 expected no response
ok 21 cluster has an empty state
# relay request handles channel close correctly
TCHANNEL WARN: dropping error from dead relay request ~ { error:
   { [TchannelLocalResetError: tchannel: tchannel: Connection was manually closed.]
     type: 'tchannel.local.reset',
...
```

relay_handler.js
```diff
@@ -221,23 +221,16 @@ function relayRequestlogError(err, codeName) {
     });
 };
 
-LazyRelayInReq.prototype.checkTimeout =
```
Raynos: where did `checkTimeout` go ?

Raynos:
Looks good enough.

relay_handler.js
jcorbin: it's now onTimeout, checkTimeout was never right for these classes.

kriskowal:
:ok:

PR #38, opened 2015-10-06
[SEMVER-MAJOR]: hyperbahn-client: ban the ability to create a client channel for yourself

Raynos:
This change cleans up an edge case that was previously confusing.

Now when you try and create a duplicate sub channel it simply tells you
to go away.

This reduces confusion around whether requestDefaults and the like
are set properly.


Raynos:
RIP

PR #40, opened 2015-10-06
Fix lingering peer options anti-pattern

kriskowal:


kriskowal:
r @jcorbin

Raynos:
lgtm.

Raynos:
#yolo

PR #41, opened 2015-10-07
only emit relay.latency after writing bytes to socket

Raynos:
We spend non-trivial CPU in the entire bufrw and net.Socket
layer converting call response into bytes.

We need to capture that in this stat

r: @rf @jcorbin @kriskowal


jcorbin:
lgtm once past tests

PR #42, opened 2015-10-07
[as-http] Remove request body size limit

anson627:
It turns out that request body can be larger than 1 mb, so remove the limit to avoid getting 413 error code

@vipulaneja  @jcorbin @ShanniLi


Raynos:
You want a limit. There is only so much RAM available.

If you have 2GB of RAM & 1k concurrent requests you should set the limit to 20mb or do streaming.

Honestly :-1: keep the 1mb limit and implement streaming for anything over the limit.

anson627:
20mb sounds good, we will figure out streaming later on.

ShanniLi:
:+1:

PR #46, opened 2015-10-08
update Meta::thriftIDL

ShanniLi:
r @Raynos @kriskowal 
cc @junchaowu @prashantv @blampe

### DIFF
--- a/as/meta.thrift
+++ b/as/meta.thrift
@@ -3,7 +3,14 @@ struct HealthStatus {
     2: optional string message
 }
 
+struct ThriftIDLs {
+    // map: filename -> contents
+    1: required map<string, string> idls
+    // the entry IDL that imports others
+    2: required string entryPoint
+}
+
 service Meta {
     HealthStatus health()
-    string thriftIDL()
+    ThriftIDLs thriftIDL()
 }
--- a/as/meta.thrift
+++ b/as/meta.thrift
@@ -3,11 +3,13 @@ struct HealthStatus {
     2: optional string message
 }
 
+typedef string filename
+
 struct ThriftIDLs {
     // map: filename -> contents
-    1: required map<string, string> idls
+    1: required map<filename, string> idls
     // the entry IDL that imports others
-    2: required string entryPoint
+    2: required filename entryPoint
 }
 
 service Meta {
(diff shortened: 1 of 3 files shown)


as/meta.thrift
```diff
@@ -3,7 +3,14 @@ struct HealthStatus {
     2: optional string message
 }
 
+struct ThriftIDLs {
+    // map: filename -> contents
+    1: required map<string, string> idls
+    // the entry IDL that imports others
+    2: required string entryPoint
```
kriskowal: consider `typedef filename string` and using the filename type here to disambiguate the IDL for the humans.

kriskowal:
lgtm

as/thrift.js
```diff
@@ -45,6 +45,7 @@ function TChannelAsThrift(opts) {
         'must pass source as an argument');
 
     self.thriftSource = opts.source;
+    self.thriftFileName = opts.thriftFileName || 'service';
```
kriskowal: consider `service.thrift` as the default.

as/meta.thrift
ShanniLi: Good point. Need to land this first:
https://github.com/uber/tchannel/pull/1390

as/thrift.js
ShanniLi: Not sure like the idea of having a dot in the map key. Plus, it should always be ".thrift'. Will make the change if got a '+1' on 'service.thrift'. :)

ShanniLi: Actually, I am convinced. Will make the change.

as/thrift.js
```diff
@@ -45,6 +45,7 @@ function TChannelAsThrift(opts) {
         'must pass source as an argument');
 
     self.thriftSource = opts.source;
+    self.thriftFileName = opts.thriftFileName || 'service.thrift';
```
Raynos: nope.

This argument must either be undefined or not.

We should only implement `thriftIDL` if this is specified.

Raynos: Also please update https://github.com/uber/tchannel-node/blob/master/hyperbahn-client.js#L179-L197 to pass argument.

kriskowal: We can follow-up to support AsThrift({idls, entryPoint}) when ThriftRW supports that form of constructor. For now, this is a fine migration step.

ShanniLi: oops, click too fast ... Will create a separate PR ...

PR #47, opened 2015-10-08
only implement Meta::thriftIDL when the thriftFileName is provided

ShanniLi:
r @Raynos @kriskowal 

ref: #46

### DIFF
--- a/hyperbahn-client.js
+++ b/hyperbahn-client.js
@@ -22,6 +22,7 @@
 
 var assert = require('assert');
 var fs = require('fs');
+var path = require('path');
 var TypedError = require('error/typed');
 var WrappedError = require('error/wrapped');
 var timers = require('timers');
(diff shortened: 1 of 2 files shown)


hyperbahn-client.js
```diff
@@ -188,6 +188,7 @@ function getThriftSync(options) {
 
     return channel.TChannelAsThrift({
         source: thriftSource,
+        thriftFileName: options.thriftFile.split("/").pop(),
```
kriskowal: https://nodejs.org/api/path.html#path_path_basename_p_ext

kriskowal:
@Raynos I find this change strictly less useful and will delay useful outcomes for tcurl.

ShanniLi:
+1 for @kriskowal

kriskowal:
:ok:

PR #48, opened 2015-10-08
Test and fix v2.CallResponse.RW.lazy.readArg1

jcorbin:
Never worked, viva la test coverage.

r @Raynos @kriskowal


kriskowal:
:ok:

Raynos:
lgtm.

PR #49, opened 2015-10-08
Add lazy v2 call header reading

jcorbin:
So that the upcoming stats path can get such things as the `cn` header.

r @Raynos @kriskowal 

Cheers @abhinav for making me think to use a typed array.

### DIFF
--- a/v2/header.js
+++ b/v2/header.js
@@ -192,6 +192,36 @@ HeaderRW.prototype.readFrom = function readFrom(buffer, offset) {
     return bufrw.ReadResult.just(offset, headers);
 };
 
+HeaderRW.prototype.lazyRead = function lazyRead(frame, offset) {
+    var self = this;
+
+    // TODO: conspire with Call(Request,Response) to memoize headers start/end
+    // offsets, maybe even start of each key?
+
+    var res = self.countrw.readFrom(frame.buffer, offset);
+    if (res.err) return res;
+    offset = res.offset;
+
+    var keyvals = new KeyVals(frame.buffer, res.value);
+    for (var i = 0; i < keyvals.length; i++) {
+        res = self.keyrw.sizerw.readFrom(frame.buffer, offset);
+        if (res.err) return res;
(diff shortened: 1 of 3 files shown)


v2/header.js
```diff
@@ -229,3 +259,59 @@ module.exports.header2 = HeaderRW(bufrw.UInt16BE, bufrw.str2, bufrw.str2, {
...
+    if (this.index < this.data.length) {
+        this.data[this.index++] = keyOffset;
+        this.data[this.index++] = keyLength;
+        this.data[this.index++] = valOffset;
+        this.data[this.index++] = valLength;
+    }
+};
+
+KeyVals.prototype.getValue =
+function getValue(key) {
```
kriskowal: aka get

v2/header.js
```diff
@@ -229,3 +259,59 @@ module.exports.header2 = HeaderRW(bufrw.UInt16BE, bufrw.str2, bufrw.str2, {
     maxHeaderCount: Infinity,
     maxKeyLength: Infinity
 });
+
+function KeyVals(buffer, length) {
```
kriskowal: aka lazymap

kriskowal:
:ok:

v2/header.js
```diff
@@ -229,3 +259,59 @@ module.exports.header2 = HeaderRW(bufrw.UInt16BE, bufrw.str2, bufrw.str2, {
...
+        }
+
+        var keyOffset = this.data[i];
+        var found = true;
+        for (
+            var j = 0, offset = keyOffset;
+            j < keyLength;
+            j++, offset++
+        ) {
+            if (key[j] !== this.buffer[offset]) {
```
Raynos: How much sad when using arbitrary UTF8 surrogate pairs ?

v2/header.js
Raynos: can you write a block comment on how this data structure works ?

Would be nice for future readers.

Or a TODO.

PR #51, opened 2015-10-08
v2/error_response: fix CodeNames to use idiomatic strings

jcorbin:
This collection was only used by:
- the writeGuard to validate the outgoing code (only used for existence, string value irrelevant)
- `TChannelConnection#onErrorFrame` to log unhandled error frames; low impact, log field only

r @kriskowal


kriskowal:
:ok:

PR #54, opened 2015-10-08
LazyRelayInReq: read and carry headers and endpoint

jcorbin:
r @Raynos @kriskowal

### DIFF
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -28,6 +28,9 @@ RelayHandler.RelayRequest = RelayRequest;
 
 module.exports = RelayHandler;
 
+// "constant" byte buffer used for lookup in LazyRelayInReq#initRead
+var cnBytes = Buffer('cn');
+
 function RelayHandler(channel, circuits) {
     var self = this;
     self.channel = channel;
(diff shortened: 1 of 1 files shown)


relay_handler.js
```diff
@@ -165,21 +167,31 @@ LazyRelayInReq.prototype.initRead =
...
         return res.err;
     }
     self.serviceName = res.value;
 
-    // TODO: lazy read self.callerName
+    res = self.reqFrame.bodyRW.lazy.readHeaders(self.reqFrame);
+    if (res.err) {
+        return res.err;
+    }
+    self.headers = res.value;
```
Raynos: `self.callerName = self.headers.cn` ?

relay_handler.js
```diff
@@ -205,6 +217,8 @@ function _extendLogInfo(info) {
     info.inRemoteAddr = self.remoteAddr;
     info.inRequestId = self.id;
     info.serviceName = self.serviceName;
+    info.callerName = self.callerName;
```
Raynos: no `self.callerName` yet. always undefined

relay_handler.js
jcorbin: good catch, but that's not how this headers object works, see lazyRead in v2/header for more; fix pushed to this branch to actually use `self.headers.getValue(Buffer('cn'))`

relay_handler.js
jcorbin: empty string, not undefined, but valid point still :-)

relay_handler.js
```diff
@@ -28,6 +28,9 @@ RelayHandler.RelayRequest = RelayRequest;
 
 module.exports = RelayHandler;
 
+// "constant" byte buffer used for lookup in LazyRelayInReq#initRead
+var cnBytes = Buffer('cn');
```
Raynos: nitpick: `new Buffer('cn');`

Raynos:
lgtm.

PR #55, opened 2015-10-09
Implement new circuit hooks

Raynos:
Here we manually call into the circuit hooks in the
right places (tm) for go fast and less closures.

See buddy PR because `npm link` and code organization :(
https://github.com/uber/hyperbahn/pull/38

r: @jcorbin @kriskowal

### DIFF
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -101,7 +101,7 @@ RelayHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
     // }
 
     if (self.circuits) {
-        var result = self.circuits.getCircuitForReq(req);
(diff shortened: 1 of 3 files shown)


relay_handler.js
```diff
@@ -106,6 +86,19 @@ RelayHandler.prototype._handleRequest = function _handleRequest(req, buildRes) {
     //     return;
     // }
 
+    if (self.circuits) {
+        var result = self.circuits.getCircuitForReq(req);
```
kriskowal: Please spell out Request in a method name.

Raynos: will do.

kriskowal:
:ok:

PR #57, opened 2015-10-09
Better relay stats test

jcorbin:
- reduce order sensitivity
- provide better test case description (full key path)
- harden against undefined values
- lint comply quoting
- factor out validate1 for future extensibility

r @Raynos @rf


Raynos:
#doit

PR #58, opened 2015-10-09
Add Lazy Relay Stats

jcorbin:
r @kriskowal @Raynos

### DIFF
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -377,8 +377,6 @@ function handleFrameLazily(frame) {
     // - v2.Types.CallRequestCont
     var self = this;
 
-    var now = self.channel.timers.now();
-
     if (!self.alive) {
         self.logger.warn('dropping frame from dead relay request', self.extendLogInfo({}));
         return;
@@ -394,6 +392,7 @@ function handleFrameLazily(frame) {
         }));
     }
 
+    var now = self.channel.timers.now();
     if (frame.type === v2.Types.CallRequest) {
         self._observeCallReqFrame(frame, now);
     } else if (frame.type === v2.Types.CallRequestCont) {
@@ -574,8 +573,6 @@ function handleFrameLazily(frame) {
     // - v2.Types.ErrorResponse
     var self = this;
 
-    var now = self.channel.timers.now();
-
     frame.setId(self.inreq.id);
     self.inreq.conn.socket.write(frame.buffer);
     if (frame.bodyRW.lazy.isFrameTerminal(frame)) {
@@ -585,6 +582,7 @@ function handleFrameLazily(frame) {
         }));
     }
 
+    var now = self.channel.timers.now();
     if (frame.type === v2.Types.CallResponse) {
         self._observeCallResFrame(frame, now);
     } else if (frame.type === v2.Types.CallResponseCont) {
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -66,6 +66,17 @@ RelayHandler.prototype.handleLazily = function handleLazily(conn, reqFrame) {
         return false;
     }
 
+    self.channel.emitFastStat(self.channel.buildStat(
+        'tchannel.inbound.calls.recvd',
+        'counter',
+        1,
+        new stat.InboundCallsRecvdTags(
+            rereq.callerName,
+            rereq.serviceName,
+            rereq.endpoint
+        )
+    ));
+
     conn.ops.addInReq(rereq);
     rereq.createOutRequest();
     return true;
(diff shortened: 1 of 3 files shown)


relay_handler.js
```diff
@@ -372,6 +392,78 @@ function handleFrameLazily(frame) {
...
+        'counter',
+        frame.size,
+        new stat.InboundRequestSizeTags(
+            self.callerName,
+            self.serviceName,
+            self.endpoint
+        )
+    ));
+
+    self.channel.emitFastStat(self.channel.buildStat(
```
Raynos: TODO: No garantuee that we send the frame out succesfully. frame write could fail.

relay_handler.js
```diff
@@ -372,6 +392,78 @@ function handleFrameLazily(frame) {
...
+        'counter',
+        frame.size,
+        new stat.OutboundRequestSizeTags(
+            self.serviceName,
+            self.callerName,
+            self.endpoint
+        )
+    ));
+
+    self.channel.emitFastStat(self.channel.buildStat(
```
Raynos: TODO: No garantuee that we send the frame out succesfully. frame write could fail.

relay_handler.js
```diff
@@ -372,6 +392,78 @@ function handleFrameLazily(frame) {
...
+        'counter',
+        frame.size,
+        new stat.InboundRequestSizeTags(
+            self.callerName,
+            self.serviceName,
+            self.endpoint
+        )
+    ));
+
+    self.channel.emitFastStat(self.channel.buildStat(
```
Raynos: TODO: No garantuee that we send the frame out succesfully. frame write could fail.

relay_handler.js
```diff
@@ -66,6 +66,17 @@ RelayHandler.prototype.handleLazily = function handleLazily(conn, reqFrame) {
         return false;
     }
 
+    self.channel.emitFastStat(self.channel.buildStat(
+        'tchannel.inbound.calls.recvd',
```
Raynos: This must be for CallReq only; not CallReqCont

Raynos: Ok i get it. maybe rename `handleLazily` to `handleLazyCallReq`

relay_handler.js
```diff
@@ -299,6 +310,13 @@ function onIdentified(err) {
     conn.ops.addOutReq(self.outreq);
     self.handleFrameLazily(self.reqFrame);
     self.reqFrame = null;
+
+    self.channel.emitFastStat(self.channel.buildStat(
+        'tchannel.relay.latency',
+        'timing',
+        self.outreq.start - self.start,
```
Raynos: We want `Date.now() - self.start`

relay_handler.js
```diff
@@ -358,6 +376,8 @@ function handleFrameLazily(frame) {
     // - v2.Types.CallRequestCont
     var self = this;
 
+    var now = self.channel.timers.now();
```
Raynos: I'm dubious of this `now` pattern.

I feel like `self.start` / `self.end` etc is better pattern.

relay_handler.js
```diff
@@ -421,6 +513,35 @@ LazyRelayOutReq.prototype.emitError =
...
+        new stat.OutboundCallsPerAttemptLatencyTags(
+            self.inreq.serviceName,
+            self.inreq.callerName,
+            self.inreq.endpoint,
+            self.remoteAddr,
+            0
+        )
+    ));
+
+    self.channel.emitFastStat(self.channel.buildStat(
```
Raynos: TODO: double check this is never an error frame ?

relay_handler.js
```diff
@@ -464,12 +583,32 @@ function handleFrameLazily(frame) {
...
-LazyRelayOutReq.prototype.logForwardedError =
-function logForwardedError(errFrame) {
+LazyRelayOutReq.prototype._observeErrorFrame =
+function _observeErrorFrame(errFrame, now) {
     var self = this;
 
+    self.channel.emitFastStat(self.channel.buildStat(
+        'tchannel.inbound.calls.latency',
+        'timing',
+        now - self.inreq.start,
```
Raynos: This looks like cheating to me. We should include the time for `socket.write()` in this.

`Date.now()`

relay_handler.js
```diff
@@ -498,6 +663,144 @@ function logForwardedError(errFrame) {
...
+            self.inreq.serviceName,
+            self.inreq.callerName,
+            self.inreq.endpoint
+        )
+    ));
+
+    self.channel.emitFastStat(self.channel.buildStat(
+        'tchannel.inbound.calls.latency',
+        'timing',
+        now - self.inreq.start,
```
Raynos: This is once more cheating. Need to do a `Date.now()` here to include `socket.write()` time.

Raynos:
Finished review.

Need a second opinion from @rf

relay_handler.js
jcorbin: Not a thing that the old path guaranteed either; same reply applies to all subsequent instances.

relay_handler.js
jcorbin: fair point, should be renamed, in another pr perhaps

relay_handler.js
jcorbin: good point, will do

relay_handler.js
jcorbin: we don't have self.end on either of the lazy reqs; I'm already unhappy with how many fields they've gained.

relay_handler.js
jcorbin: It's not, this entire code path (`LazyRelay{In,Out}Req`) will never normally see an error frame error object, the two normal callers are:
- LazyRealyOutReq#onTimeout
- TChannelConnection#resetAll

The only other call is TChannelConnection#onCallErrorFrame, this could only happen during a transition time (when we flip off lazy handling e.g.).  I'd instead rather add a branch to this function that makes it call `req.handleLazily` if such function exists, or similar, thoughts?

relay_handler.js
jcorbin: will do: will move down the now call until after the socket write until right before `_observe*` call tree

relay_handler.js
jcorbin: same as last, will fix

jcorbin:
Addressed feedback; the socket write "guarantee" stuff sounds interesting, but it's not a feature the old path had. It may be easier to get out of this new path tho, but I feel should be deferred to future work.

Raynos:
Lgtm. Schip it

PR #59, opened 2015-10-09
Fix uncaught exception on request without connection

kriskowal:
This has caused multiple uncaught exception in production since yesterday. The property in question is redundant with a property by the same name added by connection.extendLogInfo, on the condition that a connection is affiliated with the request.


kriskowal:
r @jcorbin

Raynos:
lgtm.

PR #60, opened 2015-10-09
bug fix: null access when connection is not availabe in req

ShanniLi:
r @kriskowal @Raynos 

ref:
https://github.com/uber/hyperbahn/issues/40
https://github.com/uber/hyperbahn/issues/41


Raynos:
duplicate.

Raynos:
https://github.com/uber/tchannel-node/pull/59

PR #61, opened 2015-10-09
Add ability to toggle lazy relaying separately from lazy handling

jcorbin:
Problem:
- we can turn lazy relaying on without incident
- but we cannot turn it off without a low defect rate

Why:
- lazy handling gets flipped only at the frame reading level currently
- however there are on going lazy requests at the time of flip off
- flipping immediately back to "eager only" frames breaks the ongoing handling of those requests

Solution:
- [x] introduce a separate control point within `RelayHandler#handleLazily` that allows us to stop creating new lazy relay requests
- [ ] once there are no more ongoing lazy relay requests, we can then turn off lazy frame reading

cc @Raynos @kriskowal @rf

### DIFF
--- a/channel.js
+++ b/channel.js
@@ -117,6 +117,7 @@ function TChannel(options) {
 
     self.options = extend({
         useLazyHandling: false,
+        useLazyRelaying: false,
         timeoutCheckInterval: 100,
         timeoutFuzz: 100,
         connectionStalePeriod: CONN_STALE_PERIOD,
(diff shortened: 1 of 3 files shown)


jcorbin:
Questions include:
- is this undue coupling between channel and relay_handler? (I for one have my doubts)
- what's the best way to get the "once all lazy relay requests are done" trigger?

Raynos:
when tests pass then shipit

rf:
tests pass locally

PR #66, opened 2015-10-12
Faster lazy relaying by deferring observability work

jcorbin:
Dear reviewers: please expand diff and consider L61, then tell me if I should add another call-site to `rereq.observabilityRead` so that the `rereq.extendLogInfo` and any stat emissions (under `rereq.sendErrorFrame`) have a chance to contain expected fields like callerName and endpoint


relay_handler.js
```diff
@@ -63,12 +63,25 @@ RelayHandler.prototype.handleLazily = function handleLazily(conn, reqFrame) {
...
-            serviceName: self.channel.serviceName
-        });
+        self.logger.warn(
+            'circuit breaking for lazy relaying isn\'t implemented',
+            rereq.extendLogInfo({}));
         return false;
     }
 
+    conn.ops.addInReq(rereq);
+    rereq.createOutRequest();
```
Raynos: what else in here will break if we delay endpoint & callerName ?

relay_handler.js
```diff
@@ -83,6 +85,8 @@ RelayHandler.prototype.handleLazily = function handleLazily(conn, reqFrame) {
         )
     ));
 
+    rereq.reqFrame = null;
```
rf: Something is breaking because this is nulled too early..

PR #67, opened 2015-10-12
more stats to debug & handle the responseAlreadyDone issue

ShanniLi:
r @Raynos @kriskowal 

FYI. this is for debugging the issue of ResponseAlreadyDone & streaming related memory leaks.

### DIFF
--- a/out_response.js
+++ b/out_response.js
@@ -213,8 +213,7 @@ TChannelOutResponse.prototype.responseAlreadyDone = function responseAlreadyDone
             callerName: self.inreq ? self.inreq.callerName : 'NA',
             serviceName: self.inreq ? self.inreq.serviceName : 'NA',
             listener: listener,
-            listeners: listeners,
-            responseObject: self
+            listeners: listeners
         });
         return true;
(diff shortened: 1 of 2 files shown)


out_response.js
```diff
@@ -195,6 +195,32 @@ TChannelOutResponse.prototype.sendCallResponseContFrame = function sendCallRespo
...
+
+    if (self.state === States.Done || self.state === States.Error) {
+        var errorEvent = self.errorEvent || {};
+        var listener = !!errorEvent.listener;
+        var listeners = 0;
+        if (errorEvent.listeners && errorEvent.listeners.length) {
+            listeners = errorEvent.listeners.length;
+        }
+
+        self.logger.error('responseAlreadyDone detected!!', {
```
Raynos: downgrade to `warn` if `codeString === Timeout`

Its possible for in requests to time out before we've finished parsing the body.

ShanniLi:
talked to @Raynos offline. This only happens before the memory blow up, so we want it to be error.

PR #68, opened 2015-10-13
multi-frame send should support non-English characters

ShanniLi:
r @jwolski @Raynos @kriskowal 
cc: @jcorbin 
ref: https://github.com/uber/ringpop-node/pull/200


Raynos:
Lgtm

PR #69, opened 2015-10-13
allow client start delay to be configurable

Raynos:
This makes it easier to start a more non trivial hyperbahn relay
that blocks on ringpop joins

r: @jcorbin @kriskowal


jcorbin:
Where's the configurable? seems like you should add an option and wire it up to an argv while we're at itS

Raynos:
@jcorbin It's "configurable" by sub classing and overwriting; That's what hyperbahn does.

There's no actual CLI config. I could add a CLI but meh.

PR #70, opened 2015-10-13
Minor Cleanups in Lazy Relaying

jcorbin:
Towards deferring observability reads until after frame relay:
- fixed a typo in a log
- added missing log context to said log
- defer counting the received call until after frame relay
- drop a `Date.now` call whose result wasn't used
- stop carrying headers on `LazyRelayInReq` as it holds a reference to the req frame buffer

r @rf @Raynos


rf:
:+1:

PR #71, opened 2015-10-13
v2/args: fix subtlety in recent args string fix

jcorbin:
Out of #68:
- this length check wasn't just not needed...
- ...it's also potentially harmful since if the arg is long enough to overflow a frame, it will continue to not get chunked correctly

r @ShanniLi @Raynos


ShanniLi:
:+1:

Raynos:
lgtm.

PR #73, opened 2015-10-13
& only works on numbers

Raynos:
We forgot to unwrap the Result object which broke
all fragmented things in production.

r: @rf @jcorbin

### DIFF
--- a/v2/call.js
+++ b/v2/call.js
@@ -157,7 +157,7 @@ CallRequest.RW.lazy.readArg1 = function readArg1(frame, headers) {
 
 CallRequest.RW.lazy.isFrameTerminal = function isFrameTerminal(frame) {
     var flags = CallRequest.RW.lazy.readFlags(frame);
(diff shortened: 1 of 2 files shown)


rf:
sure :+1:

prashantv:
Can we add tests to catch this before getting to production?

jcorbin:
We do have tests prashant:
- the entire test suite, if run with lazy relaying on
- the hyperbahn integration tests, if run with lazy relaying on

We're going to work towards doing full test-suite testing with different defaults so that we can more thoroughly test flagged-features, especially new ones that default off like this one.  Eventually such test will morph to be regression tests against the "what if we turn this feature off" case.

jcorbin:
@prashantv here's the support promised: https://github.com/uber/tchannel-node/pull/74 then we just need to use it for various config permutations.

prashantv:
Awesome, thanks @jcorbin

PR #75, opened 2015-10-14
lazy circuits

rf:
r @Raynos @jcorbin 

This is a working PR for lazy circuits. The way I'm handling the connection errors is a little jank right now; `LazyInReq.alive` should probably be a tri-state, so I'm going to try and clean that up first.

TODO also is make alloc cluster respect a config flag that will enable lazy* rather than just putting those lines in manually.

### DIFF
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -644,7 +644,7 @@ function _observeErrorFrame(errFrame, now) {
     var code = res.value;
     var codeName = v2.ErrorResponse.CodeNames[code] || 'unknown';
 
-    if (self.inreq && self.inreq.circuit) {
+    if (self.inreq.circuit) {
         if (errors.isUnhealthy(codeName)) {
             self.inreq.circuit.state.onRequestUnhealthy();
         } else {
@@ -746,7 +746,7 @@ function _observeCallResFrame(frame, now) {
         )
     ));
 
-    if (self.inreq && self.inreq.circuit) {
+    if (self.inreq.circuit) {
         self.inreq.circuit.state.onRequestHealthy();
     }
 
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -109,16 +109,15 @@ RelayHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
     // }
 
     if (self.circuits) {
-        var result = self.circuits.getCircuitForRequest(req);
-        if (result.err) {
-            var errFrame = result.err;
-            buildRes().sendError(errFrame.codeName, errFrame.message);
+        var circuit = self.circuits.getCircuit(req.headers.cn || 'no-cn', req.serviceName, String(req.arg1));
+        if (!circuit.state.shouldRequest()) {
+            buildRes().sendError('Declined', 'Service is not healthy');
             return;
         }
 
-        var circuit = result.value;
         req.circuit = circuit;
         circuit.state.onRequest(req);
+
     }
 
     req.forwardTrace = true;
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -63,10 +63,14 @@ RelayHandler.prototype.handleLazily = function handleLazily(conn, reqFrame) {
     }
 
     if (self.circuits) {
-        self.logger.warn(
-            'circuit breaking for lazy relaying isn\'t implemented',
-            rereq.extendLogInfo({}));
-        return false;
+        var circuit = self.circuits.getCircuit(rereq.callerName, rereq.serviceName, rereq.endpoint);
+        if (!circuit.state.shouldRequest()) {
+            rereq.sendErrorFrame('Declined', 'Service is not healthy');
+            return true;
+        }
+
+        rereq.circuit = circuit;
+        circuit.state.onRequest(rereq);
     }
 
     conn.ops.addInReq(rereq);
@@ -155,6 +159,8 @@ function LazyRelayInReq(conn, reqFrame) {
     self.operations = null;
     self.timeHeapHandle = null;
     self.endpoint = '';
+    self.error = null;
+    self.tracing = null;
 
     self.boundExtendLogInfo = extendLogInfo;
     self.boundOnIdentified = onIdentified;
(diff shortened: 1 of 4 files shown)


relay_handler.js
```diff
@@ -63,10 +63,14 @@ RelayHandler.prototype.handleLazily = function handleLazily(conn, reqFrame) {
     }
 
     if (self.circuits) {
-        self.logger.warn(
-            'circuit breaking for lazy relaying isn\'t implemented',
-            rereq.extendLogInfo({}));
-        return false;
+        var circuit = self.circuits.getCircuit(rereq.callerName, rereq.serviceName, rereq.endpoint);
```
Raynos: getCircuitForRequest ?

Raynos:
```
   if (self.circuits) {		     if (self.circuits) {
         var result = self.circuits.getCircuitForRequest(req);		         var result = self.circuits.getCircuitForRequest(req);
         if (result.err) {		         if (result.err) {
             var errFrame = result.err;		             var errFrame = result.err;
             buildRes().sendError(errFrame.codeName, errFrame.message);		             buildRes().sendError(errFrame.codeName, errFrame.message);
             return;		             return;
         }		         }
 		 
         var circuit = result.value;		         var circuit = result.value;
         req.circuit = circuit;		         req.circuit = circuit;
         circuit.state.onRequest(req);		         circuit.state.onRequest(req);
     }
```

relay_handler.js
```diff
@@ -208,6 +214,10 @@ function initRead() {
     }
     self.endpoint = String(res.value);
 
+    res = self.reqFrame.bodyRW.lazy.readTracing(self.reqFrame);
```
Raynos: This will show up in the flame; we can fix later.

test/lib/alloc-cluster.js
```diff
@@ -155,6 +155,10 @@ function allocCluster(opts) {
         chan.listen(port, host);
         cluster.channels[i] = chan;
 
+
+        chan.setLazyRelaying(true);
```
Raynos: #fierce

connection.js
```diff
@@ -692,12 +692,10 @@ InitOperation.prototype.onTimeout = function onTimeout(now) {
 };
 
 TChannelConnection.prototype.sendLazyErrorFrame =
-function sendLazyErrorFrame(reqFrame, codeString, message) {
+function sendLazyErrorFrame(id, tracing, codeString, message) {
```
jcorbin: this refactor is an easy win, please pull it out first; it follows a prior refactor across v2/handler and connection

relay_handler.js
jcorbin: that's some amazing paste-mangling... what does that code block actually say?

relay_handler.js
jcorbin: yeah... and this is in the tail of `#initRead` that I'm trying to defer in #66

relay_handler.js
```diff
@@ -625,6 +644,14 @@ function _observeErrorFrame(errFrame, now) {
     var code = res.value;
     var codeName = v2.ErrorResponse.CodeNames[code] || 'unknown';
 
+    if (self.inreq && self.inreq.circuit) {
```
jcorbin: self.inreq is not nullable, we do not have a preceding pattern of null checking it anywhere

service-name-handler.js
```diff
@@ -52,7 +54,10 @@ TChannelServiceNameHandler.prototype.handleLazily = function handleLazily(conn,
             error: res.err
         }));
         // TODO: protocol error instead?
-        conn.sendLazyErrorFrame(reqFrame, 'BadRequest', 'failed to read serviceName');
+        res = self.reqFrame.bodyRW.lazy.readTracing(self.reqFrame);
+        tracing = res.err ? v2.Tracing.emptyTracing : res.value;
+
+        conn.sendLazyErrorFrame(reqFrame.id, tracing, 'BadRequest', 'failed to read serviceName');
```
jcorbin: actually, let's have our cake and eat it too:
- add back something like `conn.sendLazyErrorFrameFor(reqFrame, code, mess)`

test/lib/alloc-cluster.js
jcorbin: yeah, this can't land as is, instead you'll want to use our new paramterized testing feature, which we need to get a lazy and un-lazy mode into travis Soon ™.

relay_handler.js
Raynos: All I'm saying is make lazyHandling and eager handling use the same `circuits` logic.

rf: I don't really have a problem with them having separate methods. `getCircuitForRequest` makes sense since eager handling will have an *actual* request object. I could make the eager path also use `getCircuit`; that would just be moving a bit of code from `hyperbahn/circuits.js` to `relay_handler.js`. It's a simple refactor that I don't think is necessary for this PR.

Raynos: I just want consistency so its easy to manually verify that eager and lazy are the same.

Let's move the circuits code out of relay_handler into service-proxy (real soon tm)

Raynos: It's a lot easier to "verify" that this PR is "correct" if you fix eager to have the same code.

rf:
Removed the lazy* alloc-cluster #fierce. The tests pass for both eager and lazy.

relay_handler.js
```diff
@@ -105,16 +109,15 @@ RelayHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
     // }
 
     if (self.circuits) {
-        var result = self.circuits.getCircuitForRequest(req);
-        if (result.err) {
-            var errFrame = result.err;
-            buildRes().sendError(errFrame.codeName, errFrame.message);
+        var circuit = self.circuits.getCircuit(req.headers.cn || 'no-cn', req.serviceName, String(req.arg1));
+        if (!circuit.state.shouldRequest()) {
```
Raynos: are you sure there are not multiple types of failure ?

rf: `getCircuitForRequest` would also send an error frame if it couldn't find a serviceName, but ServiceDispatchHandler does the same check earlier, so it was unnecessary.

Raynos:
Good enough; lgtm. #shipit

jcorbin:
Rebased on master, tests now green, lgtm

Raynos:
noice.

PR #76, opened 2015-10-14
TChannel: ensure that close callback does not early fire

jcorbin:
While chasing #31, I noticed that there was room for pre-mature callback out of `TChannel#close`.

r @abhinav


abhinav:
LGTM

PR #78, opened 2015-10-14
connection: refactor sendLazyErrorFrame into 2 methods

rf:
r @jcorbin @Raynos 

Refactor `sendLazyErrorFrame` into one that takes `id` and `tracing` and one that takes a `reqFrame`


Raynos:
lgtm. #shipit

PR #84, opened 2015-10-14
Better hyperbahn link test

jcorbin:
Now supports:
- easy local run against existing hyperbahn checkout
- copies any $TCHANNEL_TEST_CONFIG to the destination hyperbahn to avoid ENOENT problem since `npm run ...` does a chdir to the destination dir

r @rf @abhinav

### DIFF
--- a/hyperbahn_link_test.sh
+++ b/hyperbahn_link_test.sh
@@ -2,7 +2,7 @@
 set -e
 
 if [ -z "$1" ]; then
-    dest_dir=$(mktemp -d)
+    dest_dir=$(mktemp -d -t tchannel.XXXXXX)
     git clone https://github.com/uber/hyperbahn "$dest_dir"
 else
     dest_dir=$1
--- a/hyperbahn_link_test.sh
+++ b/hyperbahn_link_test.sh
@@ -1,4 +1,25 @@
 #!/bin/bash
+
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


hyperbahn_link_test.sh
```diff
@@ -0,0 +1,26 @@
...
+else
+    dest_dir=$1
+fi
+
+if [ -n "$TCHANNEL_TEST_CONFIG" ]; then
+    for part in $(eval "echo $TCHANNEL_TEST_CONFIG"); do
+        dest_part="$dest_dir/from_tchannel_$(echo "$part" | tr '/' '_')"
+        cp -fv "$part" "$dest_part"
+    done
+    TCHANNEL_TEST_CONFIG=$(echo "$TCHANNEL_TEST_CONFIG" | tr '/' '_')
```
rf: `TCHANNEL_TEST_CONFIG=from_tchannel_$(echo "$TCHANNEL_TEST_CONFIG" | tr '/' '_')` ?

hyperbahn_link_test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -e
+
+if [ -z "$1" ]; then
+    dest_dir=$(mktemp -d)
```
rf: maybe change this to something like `mktemp -d /tmp/tchannel.XXXXXXX` so it's osx compat

hyperbahn_link_test.sh
jcorbin: yup, so that we don't need a bunch of `mkdir -p $(dirname ...))`

hyperbahn_link_test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -e
+
+if [ -z "$1" ]; then
+    dest_dir=$(mktemp -d -t tchannel.XXXXXX)
+    git clone https://github.com/uber/hyperbahn "$dest_dir"
+else
+    dest_dir=$1
```
abhinav: nit:

    dest_dir="$1"

abhinav:
lgtm

PR #79, opened 2015-10-14
WIP: travis test

jcorbin:

PR #88, opened 2015-10-15
Improve server pool test

jcorbin:
Test improvement towards #79


Raynos:
lgtm.

Thanks for the cleanup.

jcorbin:
Cool, will have more to come, but this is enough of a chunk to land for now ;-)

PR #89, opened 2015-10-15
implement init req/res blocking for lazy relay

Raynos:
We have a race condition where we get a CallRequestCont frame before
we finish the init req/res handshake.

This used through an uncaught about `self.outreq.conn` being undefined
because waitForIdentified() had not finished yet.

Now we have an unbounded list of ContFrames for the id and we flush
the list either when we've flushed the CallRequest or when we call
InReq.onError()

r: @jcorbin @rf @kriskowal

### DIFF
--- a/relay_handler.js
+++ b/relay_handler.js
@@ -160,6 +160,7 @@ function LazyRelayInReq(conn, reqFrame) {
     self.endpoint = '';
     self.error = null;
     self.tracing = null;
+    self.reqContFrames = [];
 
     self.boundExtendLogInfo = extendLogInfo;
     self.boundOnIdentified = onIdentified;
@@ -311,6 +312,12 @@ function onIdentified(err) {
     self.forwardTo(conn);
 };
 
+LazyRelayInReq.prototype.freeReferences = function freeReferences() {
+    var self = this;
+
+    self.reqContFrames.length = 0;
+};
+
 LazyRelayInReq.prototype.forwardTo =
 function forwardTo(conn) {
     var self = this;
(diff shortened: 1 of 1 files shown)


rf:
test?

relay_handler.js
```diff
@@ -328,6 +335,12 @@ function forwardTo(conn) {
     self.handleFrameLazily(self.reqFrame);
     self.reqFrame = null;
 
+    for (var i = 0; i < self.reqContFrames.length; i++) {
+        self.handleFrameLazily(self.reqContFrames[i]);
+    }
```
kriskowal: perhaps consolidate this into a method simply called `flush()`

Raynos:
@rf I have a test linked in hyperbahn. No test in tchannel itself.

kriskowal:
Link test here for reference please. Would be nice to have validation in tchannel proper, but integration is satisfying.

relay_handler.js
Raynos: I'll leave it here; it's very order specific around

 - `self.reqFrame = null;
 - `for loop`
 - `array.length = 0;`

A little fragile..

Raynos:
https://github.com/uber/hyperbahn/blob/master/test/forward/forwarding-huge-buffer.js

relay_handler.js
rf: `flushContFrames()`? and put the "free" in there too.

kriskowal:
I will buy it, but encourage continued thinking about how to make this state machine less like a spaghetti machine with fragile ordering of operations in multiple code paths.

Raynos:
#

Raynos:
#STATE

PR #91, opened 2015-10-16
TChannelPeer: drop closure around connection handling

jcorbin:
Re-use the constructor closure instead.

r @Raynos @kriskowal


Raynos:
lgtm.

PR #92, opened 2015-10-16
Limit sanity sweep

jcorbin:
This makes it so that sanity sweeping cannot unboundedly block the event loop.

r @Raynos @kriskowal

### DIFF
--- a/peers_base.js
+++ b/peers_base.js
@@ -68,15 +68,28 @@ TChannelPeersBase.prototype.close = function close(peers, callback) {
     }
 };
 
-TChannelPeersBase.prototype.sanitySweep = function sanitySweep() {
+TChannelPeersBase.prototype.sanitySweep =
+function sanitySweep(callback) {
     var self = this;
 
-    var peers = self.values();
-    for (var i = 0; i < peers.length; i++) {
+    nextPeer(self.values(), 0, callback);
+
+    function nextPeer(peers, i, done) {
+        if (i >= peers.length) {
+            done(null);
+            return;
+        }
+
         var peer = peers[i];
         for (var j = 0; j < peer.connections.length; j++) {
             var conn = peer.connections[j];
             conn.ops.sanitySweep();
+        }
+
+        setImmediate(deferNextPeer);
+
+        function deferNextPeer() {
+            nextPeer(peers, i + 1, done);
         }
     }
 };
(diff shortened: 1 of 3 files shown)


peers_base.js
```diff
@@ -68,15 +68,46 @@ TChannelPeersBase.prototype.close = function close(peers, callback) {
...
 };
 
-TChannelPeersBase.prototype.sanitySweep = function sanitySweep() {
+TChannelPeersBase.prototype.sanitySweep =
+function sanitySweep(callback) {
     var self = this;
 
-    var peers = self.values();
-    for (var i = 0; i < peers.length; i++) {
+    nextPeer(self.values(), 0, callback);
```
jcorbin: Dear reviewers: please fact check my on iterating a list that may mutate while I iterate it.

peers_base.js
```diff
@@ -68,15 +68,46 @@ TChannelPeersBase.prototype.close = function close(peers, callback) {
...
+            done(null);
+            return;
+        }
+
         var peer = peers[i];
-        for (var j = 0; j < peer.connections.length; j++) {
-            var conn = peer.connections[j];
-            conn.ops.sanitySweep();
+
+        nextConn(peer.connections, 0, function connSweepDone(err) {
```
jcorbin: Dear reviewers: please fact check my on iterating a list that may mutate while I iterate it.

peers_base.js
kriskowal: Checked: values() returns a shallow copy.

peers_base.js
```diff
@@ -68,15 +68,46 @@ TChannelPeersBase.prototype.close = function close(peers, callback) {
     }
 };
 
-TChannelPeersBase.prototype.sanitySweep = function sanitySweep() {
+TChannelPeersBase.prototype.sanitySweep =
+function sanitySweep(callback) {
```
kriskowal: Please limit the number of concurrent sweepers to one.

peers_base.js
Raynos: This method is called infrequently. Just call `peers.slice()` its cheap enough.

peers_base.js
Raynos: This method is called infrequently. Just call `connections.slice()` its cheap enough.

PR #97, opened 2015-10-18
WIP: faster peer pending by incremental computation

jcorbin:
Early share, not ready for review yet.

Keeps a `pendingTotal` up to date on each peer by event-delivered deltas from each connection.

This is a related but orthogonal idea to #79 that I did to see if I could just make linear peer scoring less expensive.  The wall clock on `test/pool-of-servers` didn't budge much (there was maybe a ~1-2% gain on its ~5 second runtime...) however it may end up being an important win in production.

TODO:
- [ ] separately PR the operations pending change since it's needed for heap and this, and a big enough feature to deserve its own review
- [ ] add config modes to travis
- [ ] verify any performance impact


Raynos:
Looks like this is an optimization for chooseLinear().

If this doesn't help peer heap performance we might want to shelve this PR.

jcorbin:
It should still be a small micro optimization for heap selection, but yes I
intend to shelve for now. This pr is to record my experiment and serve as
that shelve.

PR #98, opened 2015-10-18
Operations: add pending change event

jcorbin:
This is support work that both #79 and #97 will need.

Open questions:
- it's currently an untested code path in this PR, but as you can see I have two upcoming tested code paths that use it... still should I unit test this feature?


Raynos:
I think the state bit approach would be nicer.

jcorbin:
Updated to use targeted approach in Operations rather than an over general eventemitter idiom.

Raynos:
lgtm.

PR #100, opened 2015-10-18
Add peer score debugging support, use it in test/pool-of-servers

jcorbin:
- add reason strings to all score invalidations
- peer score and choice observation events and test library
- optional hookup of new observation path in test/pool-of-servers
- optional bar plot the distribution in test/pool-of-servers


Raynos:
lgtm.

Raynos:
Test flap on lazy relay tests.

PR #104, opened 2015-10-20
FIx multi_bench so that it fails fast rather than hanging

jcorbin:
r @Raynos @kriskowal


jcorbin:
cc @jskelcy this is the source of your test woes in hyperbahn

kriskowal:
:ok:

PR #106, opened 2015-10-20
ThriftRW v2.5.1

kriskowal:
- This change comes with a relaxation in ThriftRW to allow sets to be expressed as lists on the wire to ease migration where the old bug is present,
- Relaxed the requirement in TChannelAsThrift that application errors be instances of Error.
- Thrift exception instances are no longer Error instances. The specialization was unhelpful, complicated, and slow.

### DIFF
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -206,8 +206,6 @@ function register(channel, name, opts, handle, spec) {
                     'expected response.body to exist');
 
                 if (!thriftRes.ok) {
-                    assert(isError(thriftRes.body),
-                        'not-ok body should be an error');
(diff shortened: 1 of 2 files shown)


kriskowal:
This change comes with a relaxation in ThriftRW to allow sets to be expressed as lists on the wire to ease migration where the old bug is present, and Thrift exception instances are no longer Error objects. The specialization was unhelpful, complicated, and slow.

jcorbin:
lgtm

kriskowal:
To green the tests, had to make a relaxation in asThrift. cc @Raynos

Raynos:
lgtm.

PR #108, opened 2015-10-20
Make TChannelPeer more easily inspectable

jcorbin:
Starting a new pattern of adding `#toString` and `#inspect` to our objects for easier debugging.

Makes it very easy to do things like `console.log(peers.toString())` for a short `TChannelPeer(X:Y), TChannelPeer(X1:Y2),...`.


kriskowal:
:ok:

PR #109, opened 2015-10-20
Add `peer.drain(reason, callback)`

jcorbin:
A corollary of `chan.drain(reason, callback)` for initial use in hyperbahn's partial affinity peer reaping (towards https://github.com/uber/hyperbahn/pull/66).

r @Raynos


Raynos:
Seems fine on the surface.

PR #112, opened 2015-10-21
Drop unused lru-cache

jcorbin:
Why did we have this...


Raynos:
Some feature we added & removed.

Raynos:
lgtm.

PR #113, opened 2015-10-21
TChannelPeer: fix draining for incoming connections

jcorbin:
Previously we weren't calling drain on new incoming connections.


Raynos:
lgtm.

PR #115, opened 2015-10-21
tracing: use ints instead of buffers

rf:
r @Raynos @jcorbin


test/tcollector-reporter.js
```diff
@@ -152,9 +152,9 @@ test('jsonSpanToThriftSpan', function t2(assert) {
     var mapped = TCollectorReporter.jsonSpanToThriftSpan(testSpan);
 
     assert.equals(mapped.name, testSpan.name);
-    assert.deepEquals(mapped.traceId, testSpan.traceid);
-    assert.deepEquals(mapped.id, testSpan.id);
-    assert.deepEquals(mapped.parentId, testSpan.parentid);
+    assert.deepEquals(mapped.traceId, Buffer([0, 0, 0, 0, 0, 0, 0, 1]));
```
jcorbin: why for these buffers still?

jcorbin:
I like less buffers, and you updated the tests, but I'm a bit confused why still buffers?

test/tcollector-reporter.js
Raynos: Let's use the thrift long support once its not broken.

rf: These are about to be thrift serialized, and in the thrift file these are still `binary`. `jsonSpanToThriftSpan` converts a span from the internal TChannel representation to something ready to be sent to the `tchannelThrift.send` call.

Raynos:
Please Long(low, high) instead of [high, low].

Let's add a dep on Long.

rf:
These aren't Longs, they're 64 random bits. Using Long to represent them is unnecessary complexity. Also, they're `binary` in the thrift file, not i64s.

Raynos:
I see. Hmm. stop using arrays and have double the fields XD

test/tcollector-reporter.js
Raynos: Oh the thrift is binary; I guess we have to buffers. #coolstory

Raynos:
lgtm. Ill stop being annoying.

kriskowal:
We can add a `binary (js.type = 'Long')` annotation to thriftrw if needs be.

PR #117, opened 2015-10-22
Add directional support to `peer.drain`

jcorbin:
Especially check out the tests, they essentially implement the "N + 1 connection pruning" we want in hyperbahn; one test is server side prune, other test is client side prune.

r @rf @Raynos


Raynos:
This tells a story. lgtm.

PR #120, opened 2015-10-22
Fix benchmark by using self.callback

Raynos:
The benchmark would abort way too early previously.

Not sure why this fixes it but I git bisected to unbreaking
the benchmark so #fierce

r: @jcorbin @rf

### DIFF
--- a/benchmarks/multi_bench.js
+++ b/benchmarks/multi_bench.js
@@ -110,6 +110,8 @@ Test.prototype.run = function run(callback) {
     var self = this;
     var i;
 
+    this.callback = callback;
+
     var counter = numClients;
     for (i = 0; i < numClients; i++) {
         self.newClient(i, onReady);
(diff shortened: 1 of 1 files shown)


jcorbin:
:ok: I'll accept that we can't make this more explicable without utterly refactoring Test, here's what I learned from trying to fix it just now:
- the problem lies around that fillPipeline is used both to start the test (where the final callback is passed in)
- and also used to re-fill the pipeline after any request finishes (now with a differing callback)
- however it is this "re-fill" case that can also decide that the whole test run is done, therefore it needs to call the final callback

I tried briefly to refactor sendNext to be an async loop which we could pass a callback through, but it didn't end well, altho I nearly got it working.  Really I think Test would be better done as an event emitter with a normal `error` event and something like a `done` event.  Then `.start` could do the convenience thing of hooking up its callback to these events.  As it stands, the bricolage approach used by Test is harder to understand than use of the debatable EventEmitter pattern.

Raynos:
This code is messy. I just want it to go back to "working".

We can improve later :)

PR #121, opened 2015-10-22
WIP: errors.logLevel: downgrade ProtocolError to warn

jcorbin:
This has caused a few spurious alerts so far in the lazy relaying path due to trying to read a zero ttl.

### DIFF
--- a/errors.js
+++ b/errors.js
@@ -860,12 +860,14 @@ module.exports.isFatal = function isFatal(err, codeName) {
 
 module.exports.logLevel = function errorLogLevel(err, codeName) {
     switch (codeName) {
-        case 'ProtocolError':
         case 'UnexpectedError':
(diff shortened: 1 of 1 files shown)


Raynos:
Id rather not ban all protocol errors.

We need to be forgiving when parsing and strict when serializing.

If we generate a protocol error because we wrote bad data then that's definitely a bug. 

Is there another solution here ? This is yet another form of ours vs theirs

jcorbin:
Good point, this needs rethinking

Raynos:
One approach we can take to fix on call is:

 - Add a new `parseError: true` field to the InvalidTtlError
 - Only downgrade to "warn" if `parseError === true`

This allows us to still get alerted about protocol related bugs but allows us to action on the zero alert policy.

PR #124, opened 2015-10-23
Add writeBuffer mode: ephemeral vs static

jcorbin:
r @kriskowal @rf @Raynos

### DIFF
--- a/v2/handler.js
+++ b/v2/handler.js
@@ -112,6 +112,7 @@ TChannelV2Handler.prototype.writeCopy = function writeCopy(buffer) {
 
 TChannelV2Handler.prototype.pushFrame = function pushFrame(frame) {
     var self = this;
+
     var writeBuffer = self.writeBuffer;
     var res = v2.Frame.RW.writeInto(frame, writeBuffer, 0);
     var err = res.err;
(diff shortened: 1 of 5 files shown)


v2/handler.js
```diff
@@ -112,17 +132,20 @@ TChannelV2Handler.prototype.writeCopy = function writeCopy(buffer) {
 
 TChannelV2Handler.prototype.pushFrame = function pushFrame(frame) {
     var self = this;
-    var writeBuffer = self.writeBuffer;
+
+    var writeBuffer = self.writeBuffer || new Buffer(v2.Frame.MaxSize);
```
Raynos: cant we compute length of frame instead of new Buffer(max) ?

v2/handler.js
```diff
@@ -112,17 +132,20 @@ TChannelV2Handler.prototype.writeCopy = function writeCopy(buffer) {
...
         if (!Buffer.isBuffer(err.buffer)) err.buffer = writeBuffer;
         if (typeof err.offset !== 'number') err.offset = res.offset;
         self.writeErrorEvent.emit(self, err);
-    } else {
-        var buf = writeBuffer.slice(0, res.offset);
-        self.writeCopy(buf);
+        return;
     }
+
+    var buf = writeBuffer.slice(0, res.offset);
```
Raynos: Why slice and writeCopy when we can pre-allocate the correct buffer and write it directly without copying.

kriskowal:
:ok: legit

PR #125, opened 2015-10-23
benchmarks: increase process startup delay to be more forgiving

Raynos:
I suspect that we sometimes get flaps because 500ms simply is not
enough to start a thing.

Especially with the Hyperbahn worker which can take more then 500ms
before it listens to a port.

I've bumped the default delays to 1500ms to avoid the benchmarks
from flapping on us.

r: @jcorbin @rf @kriskowal


jcorbin:
lgtm

kriskowal:
:ok:

rf:
:+1:

PR #132, opened 2015-10-26
Hookup pending change event to invalidate peer score

jcorbin:
Out of #97, some cleanups, and hooking up the event added by #98.

r @rf


Raynos:
#legit

PR #133, opened 2015-10-26
better assertion message for peer exception

Raynos:
Not having the hostPort in prod makes this really hard to
debug.

r: @jcorbin @rf

### DIFF
--- a/peer.js
+++ b/peer.js
@@ -599,7 +599,7 @@ TChannelPeer.prototype.removeConnection = function removeConnection(conn) {
 TChannelPeer.prototype.makeOutSocket = function makeOutSocket() {
     var self = this;
     var parts = self.hostPort.split(':');
-    assert(parts.length === 2, 'invalid destination');
+    assert(parts.length === 2, 'invalid destination ' + self.hostPort);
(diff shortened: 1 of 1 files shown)


peer.js
```diff
@@ -599,7 +599,7 @@ TChannelPeer.prototype.removeConnection = function removeConnection(conn) {
 TChannelPeer.prototype.makeOutSocket = function makeOutSocket() {
     var self = this;
     var parts = self.hostPort.split(':');
-    assert(parts.length === 2, 'invalid destination');
+    assert(parts.length === 2, 'invalid destination ' + self.hostPort);
```
jcorbin: ugh, why assert and throw... how about a error log? but anyhow... I guess...

jcorbin:
:+1: since practical to debug... but it seems to me that we should:
- log a hard error/fatal
- reap the peer, and disconnect form it (if any connections)

rf:
:+1:

Raynos:
This is a programmer error.

If your going to allocate a peer with a junk hostPort then you messed up.

So we fatal, none of this graceful complexity nonsense.

PR #136, opened 2015-10-27
The great BatchStats migration

Raynos:
To enable putting hyperbahn specific stats aggregation
into tchannel we want to pass in a BatchStats instance
to TChannel constructor itself.

There will be a hyperbahn PR that sub-classes BatchStats
and implements a couple of re-write / double stat rules
for special stats that have too high cardinality

This PR is all the ground work to decouple Channel and
lib/statsd to enable the work.

r: @kriskowal @jcorbin @rf


channel.js
```diff
@@ -253,14 +218,29 @@ function TChannel(options) {
...
 
-    self.statsQueue = [];
+    self.statsd = self.options.statsd;
+    self.batchStats = null;
 
     self.requestDefaults = self.options.requestDefaults ?
         new RequestDefaults(self.options.requestDefaults) : null;
 
     if (!self.topChannel) {
+        self.batchStats = self.options.batchStats;
```
jcorbin: how about no...

channel.js
```diff
@@ -253,14 +218,29 @@ function TChannel(options) {
...
+        self.batchStatsAllocated = false;
+        if (!self.batchStats) {
```
jcorbin: just `if (!self.options.batchStats)`

jcorbin: how about:
```
if (opts.foo) {
  self.foo = opts.foo:
} else {
  self.foo = default();
}
```

lib/statsd.js
```diff
@@ -20,46 +20,199 @@
...
+        'the stats must have the "app" tag');
+
+    self.statsd = options.statsd;
+    self.logger = options.logger;
+    self.timers = options.timers;
+    self.baseTags = new StatTags(options.baseTags);
+
+    self.statsQueue = [];
+    self.batchStatTimer = null;
+    self.flushStatsRecur = flushStatsRecur;
```
jcorbin: `boundFlushStats` please

jcorbin:
lgtm, mod comments, tests pass please

PR #142, opened 2015-10-28
Fix peer heap

rf:
This works and passes tests and will need many iterations before it's done but it's ready for an initial read thru

@Raynos @jcorbin

### DIFF
--- a/peer.js
+++ b/peer.js
@@ -268,7 +268,7 @@ TChannelPeer.prototype.invalidateScore = function invalidateScore(reason) {
         scores: []
     } : null;
 
-    var score = self.scoreStrategy.getScore();
+    var score = self.scoreStrategy.getScoreRange();
     for (var i = 0; i < self.heapElements.length; i++) {
         var el = self.heapElements[i];
         if (info) {
@@ -699,9 +699,30 @@ function _maybeInvalidateScore(reason) {
     }
 };
 
+TChannelPeer.prototype.getScoreRange = function getScoreRange() {
+    var self = this;
+
+    var randomRange = self.pendingWeightedRange();
+    var scoreRange = self.scoreStrategy.getScoreRange();
+    var diff = scoreRange[1] - scoreRange[0];
+
+    var newRange = [
+        randomRange[0] * diff + scoreRange[0],
+        randomRange[1] * diff + scoreRange[0]
+    ];
+
+    return newRange;
+};
+
 TChannelPeer.prototype.getScore = function getScore() {
     var self = this;
-    return self.scoreStrategy.getScore();
+    var range = self.getScoreRange();
+    var diff = range[1] - range[0];
+    var rand = self.random();
+    if (rand === 0) {
+        rand = 1;
+    }
+    return range[0] + diff * rand;
 };
 
 module.exports = TChannelPeer;
--- a/peer.js
+++ b/peer.js
@@ -67,6 +67,7 @@ function TChannelPeer(channel, hostPort, options) {
     self.boundOnConnectionError = onConnectionError;
     self.boundOnConnectionClose = onConnectionClose;
     self.boundOnPendingChange = onPendingChange;
+    self._weightedRange = null;
 
     self.reportInterval = options.reportInterval || DEFAULT_REPORT_INTERVAL;
     if (self.reportInterval > 0 && self.channel.emitConnectionMetrics) {
@@ -252,6 +253,8 @@ TChannelPeer.prototype.setScoreStrategy = function setScoreStrategy(ScoreStrateg
 
 TChannelPeer.prototype.invalidateScore = function invalidateScore(reason) {
     var self = this;
+
+    self._weightedRange = null;
 
     if (!self.heapElements.length) {
         return;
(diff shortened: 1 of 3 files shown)


peer_heap.js
```diff
@@ -20,89 +20,88 @@
...
+    assert(typeof self.random === 'function', 'PeerHeap expected random fn');
+
     self.array = [];
     // TODO: worth it to keep a tail free list like TimeHeap?
     // self.end = 0;
     self._stack = [];
 }
 
-PeerHeap.prototype.choose = function choose(threshold, filter) {
+PeerHeap.prototype.iterate = function iterate(i, fn) {
```
kriskowal: Usually spelled forEach (iterate typically returns a lazy iterator object / cursor by any name).

kriskowal:
I would like us to convince ourselves that overlapping ranges are a necessary complication cc @jcorbin

peer.js
```diff
@@ -640,11 +643,32 @@ TChannelPeer.prototype.pendingWeightedRandom = function pendingWeightedRandom()
...
+    var range = self.pendingWeightedRange();
+    var min = range[0];
+    var max = range[1];
+    var diff = max - min;
+
+    // Force rand to be (0, 1] instead of [0, 1) 
+    var rand = self.random();
+    if (rand === 0) {
+        rand = 1;
+    }
```
kriskowal: This is a smart nuance.

kriskowal:
How do you intend to validate this approach to peer selection, in terms of both performance and distributing work proportionate to the rate each peer can respond to requests?

Raynos:
Looks like valid javascript. IMMA FLAME IT.

Raynos:
- https://people.uberinternal.com/~jakev/tchannel-master-with-peer-heap.svg
 - https://people.uberinternal.com/~jakev/tchannel-master-with-random-peer.svg
 - https://people.uberinternal.com/~jakev/tchannel-master-without-peer-heap.svg

Raynos:
This peer heap is really slow; slower then linear :(

I added a random peer selection strategy for motivation in terms of WHAT the flamegraph NEEDS to look like.

peer.js
jcorbin: And separable, please separate out this fix, great find.

peer.js
```diff
@@ -253,6 +254,8 @@ TChannelPeer.prototype.setScoreStrategy = function setScoreStrategy(ScoreStrateg
 TChannelPeer.prototype.invalidateScore = function invalidateScore(reason) {
     var self = this;
 
+    self._weightedRange = null;
```
jcorbin: Originally the idea was to eagerly compute the range here everytime.

What you have instead is lazy caching.

The eager case might burn more cpu, but maybe during the less critical path(s) off of peer selection.

peer_heap.js
```diff
@@ -20,89 +20,88 @@
...
         }
     }
+
     self._stack.length = 0;
 
-    return null;
+    var chosenPeer = null;
+    var highestProbability = 0;
+    var probability;
+    for (i = 0; i < set.length; i++) {
```
jcorbin: you could inline this into the loop above without any need to keep the temporary set; doing so may well reduce the overhead of this method.

peer_heap.js
```diff
@@ -121,8 +120,9 @@ PeerHeap.prototype.clear = function clear() {
 PeerHeap.prototype.add = function add(peer) {
     var self = this;
 
-    var score = peer.scoreStrategy.getScore();
-    var i = self.push(peer, score);
+    var range = peer.pendingWeightedRange();
+
+    var i = self.push(peer, range[1]);
```
jcorbin: isn't this the end? I thought we wanted to have the start of range be the heap key?

peer_heap.js
```diff
@@ -20,89 +20,88 @@
...
-};
+    // We select into this set as long as we have ranges that overlap. Then,
+    // we introduce randomness into the set and choose
+    var seedPeer = self.array[0].peer;
+    var seedRange = seedPeer.pendingWeightedRange();
+    var set = [seedPeer];
+    var minRangeStart = seedRange[0];
+    var i;
 
-PeerHeap.prototype._chooseFilteredEl = function _chooseFilteredEl(threshold, filter) {
```
jcorbin: you got rid of code specialization for the unfiltered vs filtered case, and therefore have left performance in the abundantly common (unfiltered) case on the floor.  Please restore a specialized path for unfiltered.

peer.js
jcorbin: Also I'm pretty surprised that we had to add this field at all to the peers, the original idea was that the heapElements _are_ the cache.  I was thinking that we'd modify the heap relation/comparator to work on ranges, then the code below would be modified to rescore by calling something like `self.scoreStrategy.getRange()` rather than `.getScore()` and passing that in to each of the heapElements.

peer_heap.js
```diff
@@ -20,89 +20,88 @@
...
         var el = self.array[i];
-        if (el.score <= threshold) { // TODO: why inclusive?
-            break;
-        }
 
-        if (!filter || filter(el.peer)) {
-            return el;
+        var range = el.peer.pendingWeightedRange();
+
+        if (range[1] <= minRangeStart) {
```
jcorbin: I'm pretty sure that all of this "get the real start of the range" logic only applies to the filtered case, since in the unfiltered case it's exactly and only the heap root's.

PR #144, opened 2015-10-29
Actually retry on conn identification errors

Raynos:
We had a nice "feature" where we retried over connection
errors EXCEPT errors opening a new connection. 

This causes about 1/45 requests to fail in one of 
our production services because 1/45 hyperbahn hosts are
down.

r: @kriskowal @jcorbin @rf

cc: @blampe @prashantv Consider auditing python/go :)

cc: @MadanThangavelu This is the retry bugfix.

### DIFF
--- a/request.js
+++ b/request.js
@@ -292,7 +292,18 @@ TChannelRequest.prototype.resend = function resend() {
 
     function onIdentified(err) {
         if (err) {
-            return self.emitError(err);
+            /* emulate outReq failure */
+
+            self.outReqs.push({
+                err: err
+            });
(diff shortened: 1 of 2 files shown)


junchaowu:
lol python seems fine. We always make sure retry on successful connections. If opening connection error exists, we automatically switch to next one until run out of peers.

kriskowal:
Good catch. :eyes: on Travis, but otherwise :ok:

rf:
:+1:

jcorbin:
lgtm

PR #147, opened 2015-10-29
Add an upper bound to the lifetime of a tombstone

Raynos:
This should upper bound the number of tombstones in a process

I think long term we really should have the Operations data
structure be a LRU or some kind of fixed size upper bounded
data structure.

This should reduce GC pressure in hyperbahn

r: @jcorbin @kriskowal @rf

### DIFF
--- a/channel.js
+++ b/channel.js
@@ -1003,4 +1003,16 @@ function sanitySweep(callback) {
     self.peers.sanitySweep(callback);
 };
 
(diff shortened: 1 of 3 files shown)


rf:
cool sure :+1:

Raynos:
Test pass.

kriskowal:
I did not overhear what the plan is to support long-lived streams.

Raynos:
tombstones & streams are orthogonal features

PR #149, opened 2015-10-30
its called ops, not operations

Raynos:
Maybe I should have written a test...

r: @jcorbin @rf @kriskowal


kriskowal:
Maybe you should write a test.

:ok:

blampe:
+1 to a test :)

PR #153, opened 2015-10-30
relay_handler: be more consistent about logging

Raynos:
I removed a couple of call sites to `logError()` and replaced it
with more method calls.

Also got rid of a closure.

r: @jcorbin @rf


rf:
Cool :+1:

jcorbin:
:100:

PR #154, opened 2015-10-30
relay_handler: when logging add conneciton context

Raynos:
We should add connection context to logging on the relay.
This will greatly aid debugging.

r: @jcorbin @rf


rf:
:+1:

jcorbin:
:100:

PR #155, opened 2015-10-30
Add logging of unknown error frames.

Raynos:
This adds an extra logsite for any unknown error frames.
These generally occur in race conditions.

r: @jcorbin @rf

### DIFF
--- a/operations.js
+++ b/operations.js
@@ -226,6 +226,7 @@ Operations.prototype.getPending = function getPending() {
     return self.pending;
 };
 
+// TODO: Merge getOutTombstone() with getOutReq()
 Operations.prototype.getOutTombstone = function getOutTombstone(id) {
     var self = this;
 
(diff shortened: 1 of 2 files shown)


operations.js
```diff
@@ -226,15 +226,26 @@ Operations.prototype.getPending = function getPending() {
...
+    var self = this;
+
+    var op = self.requests.out[id] || null;
+    if (op && !op.isTombstone) {
+        return null;
+    }
+
+    return op;
+};
+
```
rf: nitpick: this is a bit of a weird pattern. Why not just have one `getOutReq` function that returns either a tombstone or a req, then check `.isTombstone` in `TChannelConnection#onCallErrorFrame`? 

I feel like this code obfuscates the flow for handling error frames when there's a tombstone.

rf:
it's fine, it works, but consider the nitpick :+1:

jcorbin:
lgtm modulo adding a TODO wrt @rf's feedback

PR #158, opened 2015-10-30
increase setTimeout to reduce flaps

Raynos:
This fixes a test flap. This test has non-deterministic
client or server timeouts.

By waiting 10ms the server is more likely to time out.

r: @jcorbin @kriskowal


jcorbin:
lgtm

PR #159, opened 2015-10-30
Improve draining test for less flaps.

Raynos:
I've finally had enough of flaps.

r: @jcorbin @kriskowal

### DIFF
--- a/test/chan_drain.js
+++ b/test/chan_drain.js
@@ -88,7 +88,7 @@ allocCluster.test('chan.drain server with a few incoming', {
     }
 
     function testdown() {
-        assert.comment('triggering drain');
+        assert.comment('--- triggering drain');
         assert.equal(finishCount, 2, 'requests have not finished');
         server.drain('testdown', drained);
         finishCount++;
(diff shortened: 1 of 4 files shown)


test/chan_drain.js
```diff
@@ -154,21 +156,20 @@ allocCluster.test('chan.drain server with a few incoming', {
     }
 });
 
-allocCluster.test('chan.drain server with a few incoming (with exempt service)', {
+allocCluster.test.only('chan.drain server with a few incoming (with exempt service)', {
```
jcorbin: only oops

test/chan_drain.js
```diff
@@ -244,14 +243,19 @@ allocCluster.test('chan.drain server with a few incoming (with exempt service)',
...
     }
 
     function drained() {
         assert.pass('drain happened');
         waitForConnRemoved(6, cluster, finish);
-        server.close(closed);
+
+        // Not garaunteed that sockets are flushed.
+        // Let's wait ~10ms
+        setTimeout(function waitUntilSocketFlush() {
```
jcorbin: y u no client.timer....

jcorbin:
:ok: , but drop the only, and unsure why you can't just use the channel timers

PR #160, opened 2015-10-31
connection: Add a log for unknown call responses

Raynos:
This adds another log where we used to be silent.

This should help debug edge cases and this log actually triggers
in the draining test.

r: @jcorbin @rf


Matt-Esch:
:+1:

PR #162, opened 2015-10-31
Fix time granularity

jcorbin:
Finally bringing this up: the time heap and its "set exactly one timer, the next one" obsoleted the need for fuzzing, let alone large interval.  For historical reasons we kept these, meaning to remove them eventually.

I'd argue that we don't even want a minimum of any sorts on the next timer, but for now I'm only proposing that we drop it to 10ms.


Raynos:
I don't understand enough about this anymore.

I would love to fact check with @mranney

Raynos:
I am convinced we should #fierce this but with a #worldclass flipr.

PR #163, opened 2015-11-02
WIP: Convert entire project form jshint -> eslint/standard

Raynos:
This is a WIP PR that does the following:

 - remove jshint
 - Add eslint rules using eslint-config-perf-standard
 - converts EVERY single file.

This is using the same linting configuration as 
https://github.com/uber/hyperbahn/pull/106

For sanity I'm adding an .eslintignore file that contains:

 - examples/*.js
 - scripts/*.js
 - test/
 - v1/*.js

This will be a first pass towards consistent linting.

The reason I'm doing this now is because it would be nice
to consistently use the `eslint` text editor plugins across
all projects related to tchannel-node/hyperbahn.

Someone else will have to make tcurl, idl and tchannel-gen comply.

r: @jcorbin @rf @kriskowal

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,7 @@ examples/*.js
 scripts/*.js
 test/
 v1/*.js
+
+# TODO remove.
+trace/*.js
+v2/*.js
--- a/connection_base.js
+++ b/connection_base.js
@@ -22,6 +22,9 @@
 
 var assert = require('assert');
 var inherits = require('util').inherits;
+var process = require('process');
+var Buffer = require('buffer').Buffer;
+
 var EventEmitter = require('./lib/event_emitter');
 var stat = require('./stat-tags.js');
 var v2 = require('./v2');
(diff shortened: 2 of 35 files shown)


Raynos:
So...

I've fixed everything EXCEPT:

 - v2/*.js ( @jcorbin )
 - trace/*.js ( @rf )

I would like to summon some #worldclass here.

image

Let's make this a team effort! If some other folks could chip in and help me migrate the rest of the production code away from jshint that would be #worldclass

.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": ["perf-standard"],
+    "rules": {
+        "max-params": [2, 5],
+        "max-len": [0],
+        "max-statements": [2, 25]
+    }
```
jcorbin: please:
```
        "no-inline-comments": [0],
        "no-multi-spaces": [0],
 ```

junchaowu:
that's patient

PR #164, opened 2015-11-02
request-handler: emit error when failed to parse arg2/arg3

Raynos:
We completely swallowed errors when failing to parse args.

This is BAD (tm).

r: @jcorbin @rf @kriskowal

### DIFF
--- a/request-handler.js
+++ b/request-handler.js
@@ -54,8 +54,12 @@ RequestCallbackHandler.prototype.handleRequest = function handleRequest(req, bui
     var res;
     if (req.streamed) {
         req.withArg23(function onArg23(err, arg2, arg3) {
+            if (err) {
+                return req.emitError(err);
+            }
+
             if (req.res && req.res.responseAlreadyDone()) {
-                return;
+                return null;
             }
 
             res = buildResponse({streamed: false});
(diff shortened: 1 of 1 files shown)


Matt-Esch:
:+1:

request-handler.js
```diff
@@ -54,8 +54,19 @@ RequestCallbackHandler.prototype.handleRequest = function handleRequest(req, bui
     var res;
     if (req.streamed) {
         req.withArg23(function onArg23(err, arg2, arg3) {
+            if (err) {
+                req.channel.logger.warn(
+                    'Could not parse arg3 for streaming inreq',
```
jcorbin: or arg2, you don't necessarily know that the problem was with arg3

request-handler.js
```diff
@@ -54,8 +54,19 @@ RequestCallbackHandler.prototype.handleRequest = function handleRequest(req, bui
...
+                    req.extendLogInfo({
+                        error: err
+                    })
+                );
+
+                return req.emitError(err);
```
jcorbin: won't it log about it? shouldn't we just wrap it here in a `arg23somethingerorr'? maybe `req#withArg23` should do such wrapping for us?

Raynos: I dont trust it; in request might have already errored and just drop the second error on the floor.

jcorbin:
:ok: once pass tests

Issue #123, opened 2015-10-23
Make :allthethings: objects inspectable


kriskowal:
Add `#toString` and `#inspect` to TChannel objects for easier debugging, à la #108 

- [ ] channel
- [x] peer
- [ ] peers
- [ ] connection (base)
- [ ] in/out req/res
- [ ] other? (discovery needed)

PR #166, opened 2015-11-02
Peer/Score strategies: ranges all the way down

rf:
~~Currently breaks pool-of-servers because it expects perfect distribution.~~

### DIFF
--- a/peer.js
+++ b/peer.js
@@ -67,6 +67,7 @@ function TChannelPeer(channel, hostPort, options) {
     self.boundOnConnectionError = onConnectionError;
     self.boundOnConnectionClose = onConnectionClose;
     self.boundOnPendingChange = onPendingChange;
+    self._range = null;
 
     self.reportInterval = options.reportInterval || DEFAULT_REPORT_INTERVAL;
     if (self.reportInterval > 0 && self.channel.emitConnectionMetrics) {
@@ -263,6 +264,8 @@ TChannelPeer.prototype.setScoreStrategy = function setScoreStrategy(ScoreStrateg
 TChannelPeer.prototype.invalidateScore = function invalidateScore(reason) {
     var self = this;
 
+    self.computeScoreRange();
+
     if (!self.heapElements.length) {
         return;
     }
@@ -275,7 +278,7 @@ TChannelPeer.prototype.invalidateScore = function invalidateScore(reason) {
         scores: []
     } : null;
 
-    var score = self.scoreStrategy.getScore();
+    var score = self.scoreStrategy.getScoreRange();
     for (var i = 0; i < self.heapElements.length; i++) {
         var el = self.heapElements[i];
         if (info) {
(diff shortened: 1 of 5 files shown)


peer.js
```diff
@@ -626,42 +629,36 @@ TChannelPeer.prototype.makeOutConnection = function makeOutConnection(socket) {
...
+// This had the attribute that a less loaded peer was merely more likely to
+// be chosen over a more loaded peer.
+// We observed with the introduction of a heap, that a less favored peer
+// would have its score less frequently re-evaluated.
+// An emergent behavior was that scores would, over time, be squeezed
+// toward zero and the least favored peer would remain the least favored
+// for ever increasing durations.
+//
+// This remains true with this algorithm, within each equivalence class.
+TChannelPeer.prototype.pendingWeightedRange = function pendingWeightedRange() {
```
jcorbin: Now With More Inlinability ™; good call.

peer.js
```diff
@@ -692,9 +689,39 @@ function _maybeInvalidateScore(reason) {
     }
 };
 
+TChannelPeer.prototype.getScoreRange = function getScoreRange() {
+    var self = this;
+
+    if (!self._range) {
```
jcorbin: Let's not add caching in this pr please, defer for later work.

Also, if we do go with this mode of caching, it should be contained in the getScoreRange method entirely:
- computeScoreRange should just compute and return
- the set of self._range should be in getScoreRange

sub_peers.js
```diff
@@ -129,7 +129,7 @@ TChannelSubPeers.prototype.chooseLinearPeer = function chooseLinearPeer(req) {
         var hostPort = hosts[i];
         var peer = self._map[hostPort];
         if (!req || !req.triedRemoteAddrs || !req.triedRemoteAddrs[hostPort]) {
-            var score = peer.scoreStrategy.getScore(req);
+            var score = peer.getScore(req);
```
jcorbin: hmm, it looks to me like we could separate out the change to the relation ship of peer and strategy from the switch to range api... thouhts?

range.js
```diff
@@ -0,0 +1,46 @@
...
+'use strict';
+
+module.exports = Range;
+
+function Range(lo, hi) {
+    var self = this;
+
+    self.lo = lo;
+    self.hi = hi;
+}
```
jcorbin: please add `#inspect` and `#toString` methods, ala #123

jcorbin:
Why does it break pool-of-servers balance? It shouldn't at all... perhaps that's because of the caching you slipped in to this one without invalidation?

rf:
No longer breaks the distribution, it turned out to just be a bug in the range logic.

rf:
Now ready for review @Raynos @jcorbin

jcorbin:
lgtm, but let's wait for @Raynos to weight in

Raynos:
#sick Seems legit.

PR #168, opened 2015-11-02
Peer: cache score ranges

rf:
r @jcorbin @Raynos 

Caches ranges on the peer object.
Ranges, no caching: https://people.uberinternal.com/~rf/ranges.svg
cached: https://people.uberinternal.com/~rf/range-cache.svg


blampe:
Is distribution still even with this?

rf:
Yes -- this caches the possible range from which we choose randomly; we don't cache the randomness.

jcorbin:
lgtm

jcorbin:
Even better, :100:

PR #170, opened 2015-11-03
Fix peer heap

rf:
r @jcorbin @Raynos 

Peer heap of ranges. Heap is organized using range max. Passes all tests and is faster than linear:

https://people.uberinternal.com/~rf/no-peer-heap.svg
https://people.uberinternal.com/~rf/peer-heap.svg

Brings peer selection from ~3.5% of process to 2.1%.


Raynos:
how about 0.5%.

How do we get there ?

peer_heap.js
```diff
@@ -24,85 +24,79 @@ module.exports = PeerHeap;
...
 
-    if (!self.array.length) {
-        return null;
-    }
-
-    var el;
-    if (filter) {
-        el = self._chooseFilteredEl(threshold, filter);
+    var chosenPeer = null;
+    var maxRangeStart = self.array[0].range.lo;
```
jcorbin: how about introduce `var el` (formerly from L61 below) here, and have `var el = self.array[0];`.

peer_heap.js
```diff
@@ -24,85 +24,79 @@ module.exports = PeerHeap;
...
     var self = this;
 
-    if (!self.array.length) {
-        return null;
-    }
-
-    var el;
-    if (filter) {
-        el = self._chooseFilteredEl(threshold, filter);
+    var chosenPeer = null;
```
jcorbin: `if (!self.array) return chosenPeer;` ?

jcorbin:
Let's update, or remove if dead, all other paths that use `el.score` to also now use/update el.range; e.g. `PeerHeapElement#rescore`.

rf:
After staring at a lot of Hydrogen IR I was able to squeeze it down a little more:
https://people.uberinternal.com/~rf/peer-heap-opt.svg

By caching the range los and his and inlining the getScore function (which just randomly chooses within the range anyway) it gets about twice as fast since they're stored as an array of unboxed doubles.

jcorbin:
Fantastic work @rf, can we get this past test sand landed soon?

peer_heap.js
```diff
@@ -40,69 +46,62 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
...
+            // This range ends before the range with the largest start begins,
+            // so it can't possibly be chosen over any of the ranges we've
+            // seen. All ranges below this one have a smaller end, so this
+            // range and any below it can't be chosen.
+            continue;
+        } else if (!filter || filter(el.peer)) {
+            maxRangeStart = Math.max(maxRangeStart, self.rangelos[i]);
+
+            // INLINE of TChannelPeer#getScore
+            var lo = self.rangelos[i];
```
Raynos: How much does this inlining help ?

Raynos: Can we at least create a function in this file and have v8 inline that function for us.

Raynos:
Let's get tests passing and then ship it.

peer_heap.js
rf: Yes, we can do that. It has a huge improvement because it removes the need for the range hi/lo property lookups. Instead, they're in a contiguous array, which improves cache locality.

Raynos: Whatever; inline that shit.

Raynos:
DO IT.

jcorbin:
:100: (please `git rm test/lib/peer_score_random.js`)

PR #171, opened 2015-11-03
Ensure we always pass a callback to sanitySweep.

Raynos:
Not passing a callback causes a uncaught exception in production.

r: @jcorbin @kriskowal @rf


jcorbin:
:100:

PR #173, opened 2015-11-04
Improve the semantics of InRequest.onTimeout()

Raynos:
There is a lot of nuanced code here that I found whilst
trying to verify the semantics of timeouts against
the relay handler.

These commits make the code stabler and get rid of
some confusing logs, namely send response after timeout
and orphaned conn onReqDone

r: @jcorbin @rf @kriskowal

### DIFF
--- a/in_request.js
+++ b/in_request.js
@@ -182,7 +182,9 @@ TChannelInRequest.prototype.onTimeout = function onTimeout(now) {
     }
 
     function deferInReqTimeoutErrorEmit() {
-        self.emitError(timeoutError);
+        if (!self.res || self.res.state === States.Initial) {
+            self.emitError(timeoutError);
+        }
     }
 };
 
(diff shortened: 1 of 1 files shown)


in_request.js
```diff
@@ -182,7 +179,14 @@ TChannelInRequest.prototype.onTimeout = function onTimeout(now) {
     }
 
     function deferInReqTimeoutErrorEmit() {
-        self.emitError(timeoutError);
+
```
jcorbin: lol extra line

time_heap.js
```diff
@@ -81,6 +83,17 @@ function TimeHeap(options) {
...
+
+    self.scheduledDrain = null;
+
+    self.boundDrainExpired = boundDrainExpired;
+
+    function boundDrainExpired() {
+        var now = self.scheduledDrain;
+        self.scheduledDrain = null;
+
+        self.drainExpired(now);
```
jcorbin: because of the order here, it's possible for `self.drainExpired` to schedule another one... is that intended?

in_request.js
Raynos: :(

time_heap.js
Raynos: @jcorbin is going to #fixit

Raynos:
I am #fierce this once test passes.

PR #175, opened 2015-11-04
Fix time heap drain

jcorbin:
Out of #173 

r @Raynos @kriskowal

### DIFF
--- a/time_heap.js
+++ b/time_heap.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+var process = require('process');
+
 var DRAIN_LOOP_MAX_DURATION = 100; // ms
 
 module.exports = TimeHeap;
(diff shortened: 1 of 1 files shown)


time_heap.js
```diff
@@ -123,6 +135,18 @@ TimeHeap.prototype.update = function update(item, now) {
...
+TimeHeap.prototype.scheduleDrainExpired =
+function scheduleDrainExpired(now) {
+    var self = this;
+
+    if (self.scheduledDrain) {
+        return;
+    }
+
+    self.scheduledDrain = now;
+    self.timers.setImmediate(self.boundDrainExpired);
```
Raynos: use the global one. mock timers does not implement `setImmediate()`

PR #178, opened 2015-11-05
Various lazy relay related cleanups.

Raynos:
All these cleanups make it easier to move the lazy relay
into a seperate file.

These are mostly about having the same semantics for lazy &
eager in the RelayHandler itself.

r: @jcorbin @rf


jcorbin:
:100: once tests green

PR #179, opened 2015-11-05
rewrite the test to be less flappppppppy

Raynos:
This test is annoying. All these drain tests are annoying

r: @jcorbin @rf


Raynos:
#fierce

PR #182, opened 2015-11-05
v2/handler: implement tchannel version init headers

jcorbin:
Implements https://github.com/uber/tchannel/pull/1398 

r @Raynos @kriskowal @rf 
cc @ShanniLi @breerly @prashantv


Raynos:
Lgtm.

PR #184, opened 2015-11-05
Clamp the maximum ttl in the relay.

Raynos:
We've had production problems where people accidentaly
set the time timeout to 1 million milliseconds (i.e. 16
minutes).

This change clamps the ttl to some sensible maximum for
safety reasons.

r: @jcorbin @rf


rf:
:+1: as soon as tests pass

PR #189, opened 2015-11-05
Add goals to peer draining

jcorbin:
Easy to accomodate now that we have the explicit `PeerDrain` object.  As
mentioned in early review for peer drain, we always wanted to have the ability
to say "and then ...".

This adds the two "and then ..."s that are used in Hyperbahn now/soon.

r @raynos @rf

### DIFF
--- a/drain.js
+++ b/drain.js
@@ -53,8 +53,8 @@ PeerDrain.prototype.extendLogInfo =
 function extendLogInfo(info) {
     var self = this;
 
-    info.drainTimeout = self.timeout;
     info.drainReason = self.reason;
+    info.drainTimeout = self.timeout;
     info.drainDirection = self.direction;
     info.drainStartedAt = self.startedAt;
     info.drainStoppedAt = self.stoppedAt;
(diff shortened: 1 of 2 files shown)


drain.js
```diff
@@ -142,6 +172,62 @@ function stop() {
...
+
+    var info = self.peer.extendLogInfo(self.extendLogInfo({
+        error: err
+    }));
+
+    if (err.type === 'tchannel.drain.peer.timed-out') {
+        self.peer.logger.warn(
+            'drain timed out, force closing connections',
+            info);
+    } else {
```
Raynos: `else if err` ?

drain.js
```diff
@@ -142,6 +172,62 @@ function stop() {
...
+
+    var info = self.peer.extendLogInfo(self.extendLogInfo({
+        error: err
+    }));
+
+    if (err.type === 'tchannel.drain.peer.timed-out') {
+        self.peer.logger.warn(
+            'drain timed out, force closing peer',
+            info);
+    } else {
```
Raynos: `else if err` ?

Raynos:
lgtm.

jcorbin:
Tested by `npm link` locally in https://github.com/uber/hyperbahn/pull/130

PR #190, opened 2015-11-06
peer heap benchmarks

rf:
The peer heap benchmarks I used to make it fast. One is for running in d8, and one uses the peer heap module directly.


Raynos:
Lgtm

PR #192, opened 2015-11-06
Fix streaming bug in as/http

Raynos:
We found a weird edgecase where we didnt send the response.

This fixes it by checking for the ended edgecase and sending
back the entire data.

r: @jcorbin @rf @Matt-Esch


rf:
ok lgtm once tests pass :+1:

PR #193, opened 2015-11-07
Adding back deferred reaping in time-heap + tests.

Raynos:
The time heap fix I made was important. I added some
failing tests that require a time heap fix.

I dont know why we leaked memory; maybe it was the
uncaught instead of the time heap change...

Let's roll forward with this change.

r: @jcorbin @rf


jcorbin:
Just as much :+1: as the first go round.

PR #195, opened 2015-11-11
Fix the double timeout problem

Raynos:
Write a test that times out on one out of 3 remotes.

Then assert about the timeouts and the logs for the timeouts.

TODO:

 - Add more asserts in test around the log lines
 - Figure out why there are no "resetting connection" logs
 - Write test for lazy relay.


Raynos:
Reminder: assert counts of req received in server

jcorbin:
:+1:

PR #197, opened 2015-11-12
allow passing host from CLI

Raynos:
This makes this program more useful because you can
listen to 127.0.0.1 if you want.

r: @jcorbin @kriskowal @rf


rf:
:+1: love it

PR #198, opened 2015-11-13
Use tombstone.extendLogInfo() when logging tombstone.

Raynos:
Not doing so gives us a cyclic object that is being logged :(

r: @jcorbin @rf


rf:
ya :+1:

PR #200, opened 2015-11-16
Check init host port strings

rf:
@Raynos 

do the thing

### DIFF
--- a/errors.js
+++ b/errors.js
@@ -242,6 +242,12 @@ Errors.InvalidHeaderTypeError = TypedError({
     name: null
 });
 
+Errors.InvalidInitHostPortError = TypedError({
+    type: 'tchannel.invalid-init-host-port',
+    message: 'invalid host:port string in init header',
+    hostPort: null
+});
+
 Errors.InvalidJSONBody = TypedError({
     type: 'tchannel-handler.json.invalid-body',
     message: 'Invalid error body, expected a typed-error',
@@ -703,6 +709,7 @@ Errors.classify = function classify(err) {
         case 'tchannel.init.send-call-response-cont-before-indentified':
         case 'tchannel.invalid-error-code':
         case 'tchannel.invalid-frame-type':
+        case 'tchannel.invalid-init-host-port':
         case 'tchannel.missing-init-header':
         case 'tchannel.protocol.invalid-ttl':
         case 'tchannel.protocol.read-failed':
--- a/v2/handler.js
+++ b/v2/handler.js
@@ -224,6 +224,13 @@ TChannelV2Handler.prototype.handleInitRequest = function handleInitRequest(reqFr
         hostPort: headers.host_port,
         processName: headers.process_name
     };
+
+    if (!validateHostPort(init.hostPort)) {
+        return self.errorEvent.emit(self, errors.InvalidInitHostPortError({
+            hostPort: init.hostPort
+        }));
+    }
+
     self.sendInitResponse(reqFrame);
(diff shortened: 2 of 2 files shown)


v2/handler.js
```diff
@@ -966,3 +971,33 @@ function InRequestOptions(
...
+    var parts = hostPort.split(':');
+    if (parts.length !== 2) {
+        return false;
+    }
+
+    if (parts[0] === '127.0.0.1') {
+        return false;
+    }
+
+    if (parts[0] === '0.0.0.0') {
```
Raynos: This breaks all kinds of things; we should check `req.remoteAddr` in advertise instead.

rf: This is the host port as passed in the header of an init request. The assertion error we were seeing was in [peer.js:566](https://github.com/uber/tchannel-node/blob/master/peer.js#L569): `assert(parts.length === 2, 'invalid destination ' + self.hostPort);`, and it's the host port used when we instantiate a peer object, which comes from the header of an init request. This happens far before the advertise path.

It's true that we allow a hostPort of `0.0.0.0:0` for ephemeral clients ([connection.js:519](https://github.com/uber/tchannel-node/blob/master/connection.js#L519)), so this particular conditional should be removed; but I do believe that this is the correct place for this check.

Raynos: @rf we need to allow 0.0.0.0 and 127.0.0.1

We can check again in hyperbahn advertise for 127.0.0.1

v2/handler.js
```diff
@@ -224,6 +224,11 @@ TChannelV2Handler.prototype.handleInitRequest = function handleInitRequest(reqFr
         hostPort: headers.host_port,
         processName: headers.process_name
     };
+
+    if (!validateHostPort(init.hostPort)) {
+        return self.errorEvent.emit(self, errors.InvalidInitHostPortError());
```
Raynos: Let's put the host port in the error message.

errors.js
```diff
@@ -242,6 +242,12 @@ Errors.InvalidHeaderTypeError = TypedError({
     name: null
 });
 
+Errors.InvalidInitHostPortError = TypedError({
+    type: 'tchannel.invalid-init-host-port',
+    message: 'invalid host:port string in init header; got: {hostPort}',
```
jcorbin: high cardinality string, leave the field up to the logging of this error object imo

v2/handler.js
```diff
@@ -966,3 +973,25 @@ function InRequestOptions(
...
     self.tracer = tracer;
 }
+
+function validateHostPort(hostPort) {
+    var parts = hostPort.split(':');
+    if (parts.length !== 2) {
+        return false;
+    }
+
+    if (parts[0].split('.').length !== 4) {
```
jcorbin: doesn't validate that the first part is a dotted quad, we need to disallow dns names; should also disallow 0.0.0.0

v2/handler.js
jcorbin: oh I see, this was asked for previously... really?

rf:
I took the hostPort out of the error message string but left it in the error object. Also validated that the host parts are all numbers.

errors.js
Raynos: we should talk about cardinality of error strings at some point.

I think there is value in it because the lack of meta data in error frames and high cardinality string in error frame is the "best" user experience we can give our customers.

Raynos:
+1 Looks good to me.

jcorbin:
ahh makes sense @raynos, lgtm then

PR #202, opened 2015-11-17
Fix lazy out request memory leak

Raynos:
I found it! It was hard to find because all of our servers
send timeout error frames for timed out in requests.

r: @jcorbin @kriskowal @rf

### DIFF
--- a/lazy_relay.js
+++ b/lazy_relay.js
@@ -546,6 +546,11 @@ LazyRelayOutReq.prototype.onTimeout =
 function onTimeout(now) {
     var self = this;
 
+    self.conn.ops.popOutReq(self.id, self.extendLogInfo({
+        info: 'lazy out request timed out',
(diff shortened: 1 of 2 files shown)


lazy_relay.js
```diff
@@ -546,6 +546,11 @@ LazyRelayOutReq.prototype.onTimeout =
 function onTimeout(now) {
     var self = this;
 
+    self.conn.ops.popOutReq(self.id, self.extendLogInfo({
+        info: 'lazy out request timed out',
+        relayDirection: 'out'
+    }));
```
kriskowal: :tada:

kriskowal:
Test case of champions. :100:

lazy_relay.js
jcorbin: I am :sob:

rf:
:+1:

PR #205, opened 2015-11-18
onTimeout always pop the out request.

Raynos:
Even if there is a response there is no garantuee that the
remote connection will ever send back an error frame or
continue frame for the out response especially if the in request
is timed out.

We need to just pop it right here

r: @jcorbin @kriskowal @rf


kriskowal:
looks right to me

PR #207, opened 2015-11-19
Updating thriftrw to 3.x

malandrew:
This is ready for a first pass review. 

The only additional thing I want to change is making sure that you need to declare that the file system will be accessed. Please confirm that this is desirable.

### DIFF
--- a/tcollector/reporter.js
+++ b/tcollector/reporter.js
@@ -21,12 +21,10 @@
 'use strict';
 
 var path = require('path');
-var fs = require('fs');
 var assert = require('assert');
 var Buffer = require('buffer').Buffer;
 
 var tcollectorSpecPath = path.join(__dirname, 'tcollector.thrift');
-var tcollectorSpec = fs.readFileSync(tcollectorSpecPath, 'utf8');
 
 module.exports = TCollectorTraceReporter;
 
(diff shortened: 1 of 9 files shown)


tcollector/reporter.js
```diff
@@ -25,8 +25,8 @@ var fs = require('fs');
 var assert = require('assert');
 var Buffer = require('buffer').Buffer;
 
-var tcollectorSpec =
-    fs.readFileSync(path.join(__dirname, 'tcollector.thrift'), 'utf8');
+var tcollectorSpecPath = path.join(__dirname, 'tcollector.thrift');
+var tcollectorSpec = fs.readFileSync(tcollectorSpecPath, 'utf8');
```
kriskowal: tcollectorSpec is no longer used.

malandrew:
I'm assuming that the default for TChannelAsThrift is to allowFileSystemAccess whereas the default for thriftrw.Thrift should be to disallow filesystem access.

kriskowal:
@malandrew Interesting observation. We could infer allowFilesystemAccess = false if options.idls is provided to the root constructor.

Looks ok. Green the tests and :ok: to land.

Raynos:
-1 breaks back compat.

Please write tests for both source and entryPoint

PR #208, opened 2015-11-21
Upgrade to thriftrw-3.0.1

malandrew:


as/thrift.js
```diff
@@ -47,8 +47,15 @@ function TChannelAsThrift(opts) {
     self.thriftSource = opts.source;
     self.thriftFileName = opts.thriftFileName || 'service.thrift';
     self.spec = new thriftrw.Thrift({
+        entryPoint: opts.entryPoint,
+        idls: opts.idls,
+        thrifts: opts.thrifts,
```
kriskowal: I don’t recall `thrifts` being an argument.

as/thrift.js
```diff
@@ -47,8 +47,15 @@ function TChannelAsThrift(opts) {
...
         source: self.thriftSource,
-        strict: opts.strict
+        strict: opts.strict,
+        allowFilesystemAccess: opts.allowFilesystemAccess,
+        allowIncludeAlias: opts.allowIncludeAlias,
+        fs: opts.fs,
+        memo: opts.memo
```
kriskowal: Let’s not expose the memo at this time. It would be harder to remove than to add when we need it.

malandrew:
@kriskowal Can I land this and update the stooge?

kriskowal:
:ok:

PR #209, opened 2015-11-22
Protect against TCP socket being full.

Raynos:
This gaurd will close the socket if we have too many
pending write operations.

This only happens if there is something wrong with the
socket; i.e. the uv_write call never calls the timeout
which we've seen in prod; something about the socket just
never calls the _write() callback.

I've checked a healthy 30% CPU worker and it only has two
pending WriteReq in memory; note that if libuv & stream_wrap
can flush the buffer synchrously then the cb to _write
gets called synchronously.

This should fix the memory leak we've seen in prod for the
last 24 hours.

r: @jcorbin @rf @kriskowal

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -235,17 +235,20 @@ function writeToSocket(buf) {
     if (self.socket._writableState.buffer.length >
         MAX_PENDING_SOCKET_WRITE_REQ
     ) {
+        var error = errors.SocketWriteFullError({
+            pendingWrites: self.socket._writableState.buffer.length
+        });
         self.logger.warn('resetting connection due to write backup',
             self.extendLogInfo({
                 pendingWrites: self.socket._writableState.buffer.length,
-                totalFastBufferBytes: self.socket._writableState.length
+                totalFastBufferBytes: self.socket._writableState.length,
+                lastBufferLength: buf.length,
+                error: error
             })
         );
 
         // NUKE THE SOCKET
-        self.resetAll(errors.SocketWriteFullError({
-            pendingWrites: self.socket._writableState.buffer.length
-        }));
+        self.resetAll(error);
         return;
     }
 
--- a/connection.js
+++ b/connection.js
@@ -33,6 +33,8 @@ var States = require('./reqres_states');
 
 var TChannelConnectionBase = require('./connection_base');
 
+var MAX_PENDING_SOCKET_WRITE_REQ = 100;
+
 function TChannelConnection(channel, socket, direction, socketRemoteAddr) {
     assert(socketRemoteAddr !== channel.hostPort,
         'refusing to create self connection'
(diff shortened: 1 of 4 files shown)


connection.js
```diff
@@ -226,6 +228,30 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
...
+        self.logger.warn('resetting connection due to write backup',
+            self.extendLogInfo({
+                bufferLength: self.socket._writableState.buffer.length,
+                maxlen: self.socket._writableState.length
+            })
+        );
+
+        // NUKE THE SOCKET
+        self.resetAll(Error.SocketWriteFullError({
+            pendingWrites: self.socket._writableState.buffer.length
```
kriskowal: pendingWrites/bufferLength to make it clear that the units are bytes, not quantity of requests, not quantity of write calls/pending frames

kriskowal:
I think we need to converse more about the dynamics of this change. Superficially, it looks like we would shed all work on a connection every time one of its sockets fills up.

Raynos:
@kriskowal if the writes times out i think we should just hard close and kill all in flight RPCs.

This is an extreme edge case and draining wont help other then wait 30 seconds.

Maybe we should have a `logger.error()` instead of `logger.warn()` for this or get better warning hygiene.

connection.js
Raynos: +1

kriskowal:
I’m informally convinced that we will not reach the pathological case because our rate limiter is configured to shed work long before we reach a request rate that exceeds our ability to respond.

If this is misconfigured, we can get to a point where we instantly go from high throughput to no throughput by crossing that threshold.

:ok:

connection.js
```diff
@@ -226,6 +228,30 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
...
+function writeToSocket(buf) {
+    var self = this;
+
+    if (self.socket._writableState.buffer.length >
+        MAX_PENDING_SOCKET_WRITE_REQ
+    ) {
+        self.logger.warn('resetting connection due to write backup',
+            self.extendLogInfo({
+                pendingWrites: self.socket._writableState.buffer.length,
+                totalFastBufferBytes: self.socket._writableState.length
```
jcorbin: `lastBufferLength: buf.length`?

connection.js
jcorbin: why don't we just create this error object above, and then use it in the warn log?

errors.js
```diff
@@ -726,6 +732,7 @@ Errors.classify = function classify(err) {
         case 'tchannel.socket':
         case 'tchannel.socket-closed':
         case 'tchannel.socket-local-closed':
+        case 'tchannel.socket.write-full':
             return 'NetworkError';
```
jcorbin: Is it? This seems more like an `UnexpectedError` to me...

PR #210, opened 2015-11-24
Defer to ThriftRW to assert source/entryPoint option pattern

kriskowal:
r @malandrew

### DIFF
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -479,14 +479,9 @@ function health(tchannelThrift, req, head, body, callback) {
 }
 
 function thriftIDL(tchannelThrift, req, head, body, callback) {
-    var idls = {};
-    idls[tchannelThrift.thriftFileName] = tchannelThrift.thriftSource;
     return callback(null, {
         ok: true,
-        body: {
-            idls: idls,
-            entryPoint: tchannelThrift.thriftFileName
-        }
+        body: tchannelThrift.spec.getSources()
     });
 }
 
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -42,7 +42,7 @@ function TChannelAsThrift(opts) {
     var self = this;
 
     self.thriftSource = opts.source;
-    self.thriftFileName = opts.thriftFileName || 'service.thrift';
+    self.thriftFileName = path.basename(opts.entryPoint || 'service.thrift');
     self.spec = new thriftrw.Thrift({
         entryPoint: opts.entryPoint,
         idls: opts.idls,
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -41,9 +41,6 @@ function TChannelAsThrift(opts) {
 
     var self = this;
 
-    assert(opts && typeof opts.source === 'string',
-        'must pass source as an argument');
-
     self.thriftSource = opts.source;
     self.thriftFileName = opts.thriftFileName || 'service.thrift';
     self.spec = new thriftrw.Thrift({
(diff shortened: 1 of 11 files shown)


malandrew:
lgtm

as/thrift.js
```diff
@@ -41,17 +41,14 @@ function TChannelAsThrift(opts) {
...
     self.thriftSource = opts.source;
-    self.thriftFileName = opts.thriftFileName || 'service.thrift';
+    self.thriftFileName = path.basename(opts.entryPoint || 'service.thrift');
     self.spec = new thriftrw.Thrift({
         entryPoint: opts.entryPoint,
         idls: opts.idls,
         source: self.thriftSource,
         strict: opts.strict,
-        allowFilesystemAccess: opts.allowFilesystemAccess,
+        allowFilesystemAccess: true,
```
malandrew: Would we ever want to allow people to disable file system access?

i.e. `opts. allowFilesystemAccess !== undefined ? opts. allowFilesystemAccess : true`

as/thrift.js
```diff
@@ -482,14 +479,9 @@ function health(tchannelThrift, req, head, body, callback) {
...
 function thriftIDL(tchannelThrift, req, head, body, callback) {
-    var idls = {};
-    idls[tchannelThrift.thriftFileName] = tchannelThrift.thriftSource;
     return callback(null, {
         ok: true,
-        body: {
-            idls: idls,
-            entryPoint: tchannelThrift.thriftFileName
-        }
+        body: tchannelThrift.spec.getSources()
```
malandrew: Weird. I thought I had already done this.

malandrew:
LGTM

as/thrift.js
kriskowal: Looks like we lost some of your prior work in #207

as/thrift.js
kriskowal: Let’s assume not for now and provide the opposite option later if needed. The only case we would disallow filesystem access for includes is when we construct Thrift from Meta::thriftIDLs, which we only intend to do in tools like tcurl/tcap.

kriskowal:
r @raynos @rf 

By way of context, I bumped the major version of ThriftRW because the semi-public interface of ThriftRW changed, although it is fully backward compatible as far as can be perceived by TChannelAsThrift. We only use the getType and getTypeResult interface of ThriftRW here, so no breaking change has been introduced.

These changes alter the tests to exercise both the TChannelAsThrift({source}) and TChannelAsThrift({entryPoint}) interfaces, by dividing the existing tests between those cases.

as/thrift.js
```diff
@@ -41,17 +41,14 @@ function TChannelAsThrift(opts) {
 
     var self = this;
 
-    assert(opts && typeof opts.source === 'string',
-        'must pass source as an argument');
-
     self.thriftSource = opts.source;
-    self.thriftFileName = opts.thriftFileName || 'service.thrift';
+    self.thriftFileName = path.basename(opts.entryPoint || 'service.thrift');
```
kriskowal: The thriftFileName is only retained here, using the basename only, to keep some tests passing that depend on it. It is not otherwise used since we call into self.spec.getSources() to produce the ThriftIDLs bundle.

malandrew:
Do we have a standard way of introducing deprecation warnings into logs so end-users are nudged in the right direction?

kriskowal:
@malandrew Yes, we would thread the logger through here. It’s too early to deprecate {source} usage though.

PR #217, opened 2015-12-01
Better listen failed error with host & port

malandrew:
This is a simple improvement to the listen failed error to make debugging a tad easier.


Raynos:
Lgtm.

PR #219, opened 2015-12-01
Support entryPoint in tchannel-as-thrift

malandrew:
Support entryPoint as well as source.

thriftSource and thriftFilename are no longer properties on tchannel-as-thrift. I don't think these should be exposed as apis. Any objection?

### DIFF
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -41,17 +41,22 @@ function TChannelAsThrift(opts) {
 
     var self = this;
 
-    self.thriftSource = opts.source;
-    self.thriftFileName = path.basename(opts.entryPoint || 'service.thrift');
+    assert(opts, 'options required');
+    assert(typeof opts === 'object', 'options must be object');
+
     self.spec = new thriftrw.Thrift({
         entryPoint: opts.entryPoint,
         idls: opts.idls,
-        source: self.thriftSource,
+        source: opts.source,
         strict: opts.strict,
         allowFilesystemAccess: true,
         allowIncludeAlias: opts.allowIncludeAlias,
         fs: opts.fs
     });
+
+    var sources = self.spec.getSources();
+    self.thriftFileName = sources.entryPoint;
+    self.thriftSource = sources.idls[sources.entryPoint];
 
     self.logger = opts.logger;
 
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "safe-json-parse": "^4.0.0",
     "sse4_crc32": "3.2.0",
     "tape-cluster": "2.1.0",
-    "thriftrw": "3.0.1",
+    "thriftrw": "^3.0.2",
     "xorshift": "^0.2.0",
     "xtend": "^4.0.0"
   },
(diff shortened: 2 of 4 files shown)


kriskowal:
@malandrew I’m reasonably certain there are no external consumers of thriftSource and thriftFilename apart from the tests, and thriftSource does not make sense in the new includes-aware world. I hesitate to remote them entirely for backward compatibility because you never know.

as/thrift.js
```diff
@@ -41,15 +41,18 @@ function TChannelAsThrift(opts) {
...
+        typeof opts.source === 'string' ||
+        typeof opts.entryPoint === 'string',
+        'must pass source as an argument'
+    );
 
-    self.thriftSource = opts.source;
-    self.thriftFileName = opts.thriftFileName || 'service.thrift';
     self.spec = new thriftrw.Thrift({
-        entryPoint: opts.entryPoint,
+        entryPoint: opts.entryPoint || opts.thriftFileName || 'service.thrift',
```
kriskowal: Is the 'service.thrift' default necessary? In the absence of an entryPoint, ThriftRW defaults the entryPoint property.

as/thrift.js
```diff
@@ -41,15 +41,18 @@ function TChannelAsThrift(opts) {
...
         idls: opts.idls,
-        source: self.thriftSource,
+        source: opts.source,
         strict: opts.strict,
         allowFilesystemAccess: opts.allowFilesystemAccess,
```
kriskowal: allowFilesystemAccess should be true in all cases, right?

malandrew:
@kriskowal should I add `thriftSource` and `thriftFilename` back in by getting them off `self.spec.getSources()` after instantiating `self.spec`?

as/thrift.js
```diff
@@ -41,12 +41,18 @@ function TChannelAsThrift(opts) {
...
 
-    self.thriftSource = opts.source;
-    self.thriftFileName = path.basename(opts.entryPoint || 'service.thrift');
+    assert(opts, 'options required');
+    assert(typeof opts === 'object', 'options must be object');
+    assert(
+        typeof opts.source === 'string' ||
+        typeof opts.entryPoint === 'string',
+        'must pass source as an argument'
+    );
```
kriskowal: We can depend on the Thrift constructor to perform this assertion.

as/thrift.js
```diff
@@ -41,12 +41,18 @@ function TChannelAsThrift(opts) {
...
+
     self.spec = new thriftrw.Thrift({
-        entryPoint: opts.entryPoint,
+        entryPoint: opts.entryPoint || 'service.thrift',
```
kriskowal: We can rely on the Thrift constructor to default the entryPoint.

kriskowal:
@malandrew You could use getSources() to provide a back-compat property. Sounds good to me.

malandrew:
Ready to land and publish.

kriskowal:
:ok:

PR #220, opened 2015-12-02
Factor out a benchmarks config module

jcorbin:
This will allow us to unify such things as remote config, command lines,
torching, taking, etc in Hyperbahn.  I'll be revving on the Makefile in
hyperbahn first, and then backing out results to tchannel.


Raynos:
legit.

PR #231, opened 2015-12-03
Update the changelog for upcoming release

jcorbin:
r @kriskowal @raynos


Raynos:
legit :)

Raynos:
We should add a line item about hostPort must be valid.

This is something that ringpop ran into.

jcorbin:
Added item about `host:port` validation

Raynos:
DO IT.

PR #232, opened 2015-12-03
Support publishing with new CHANGELOG version

jcorbin:
If the CHANGELOG has `vNEXT` as its last version, update it to the newly cut
version, and amend the fresh commit/tag to reflect.

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -28,7 +28,22 @@ if [ -z "$1" ]; then
     exit 1
 fi
 
-npm version "$1"
+if ! which tac >/dev/null 2>&1; then
+    function tac()
+    {
+        tail -r
+    }
+fi
+
+GIT_TAG=$(npm version "$1")
+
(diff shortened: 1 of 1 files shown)


publish.sh
```diff
@@ -28,7 +28,16 @@ if [ -z "$1" ]; then
...
 
-npm version "$1"
+
+GIT_TAG=$(npm version "$1")
+
+if [[ $(head -n1 CHANGELOG.md) = '# vNEXT'* ]]; then
+    sed -i -e "/^# vNEXT/c# $GIT_TAG" CHANGELOG.md
+    git commit --amend --no-edit CHANGELOG.md
+    git tag -a -f -F <(
+        git cat-file tag "$GIT_TAG"  | tac | sed -e '/^$/q' | tac | tail -n+2
```
kriskowal: tac is not on bsd. equivalent is tail -r.

jcorbin:
Made it work with OS X's userspace.

PR #233, opened 2015-12-03
break out validateHostPort into its own file

rf:
r @Raynos 

This function is useful and should be in its own file


rf:
Nope because #234

PR #234, opened 2015-12-03
Validate host port consistently in listen(), connect(), init handshake

Raynos:
This change introduces a `HostPort.validateX()` helper to allows consistent
validation of host & port concerns in all places we care.

I'll update this PR with actual markdown documentation real soon now.

@wolski asked us to clearly document what a `hostPort` actually means and
this consistent validation & documentation should help with that.

r: @rf @jcorbin

### DIFF
--- a/docs/channel.md
+++ b/docs/channel.md
@@ -93,6 +93,9 @@ Starts listening on the given port and host.
 
 Both port and host are mandatory.
 
+Port must be a valid port as per [host-port rules](./host-port.md)
+Host must be a valid non-ephemeral host as per [host-port rules](./host-port.md)
+
 The port may be 0, indicating that the operating system must grant an
 available ephemeral port.
 
--- a/docs/channel.md
+++ b/docs/channel.md
@@ -91,12 +91,10 @@ a tcollector reporter that will be wired up for you.
 
 Starts listening on the given port and host.
 
-Both port and host are mandatory.
-
-Port must be a valid port as per [host-port rules](./host-port.md)
-Host must be a valid non-ephemeral host as per [host-port rules](./host-port.md)
-
-The port may be 0, indicating that the operating system must grant an
+ - Both port and host are mandatory.
+ - Port must be a valid port as per [host-port rules](./host-port.md)
+ - Host must be a valid non-ephemeral host as per [host-port rules](./host-port.md)
+ - The port may be 0, indicating that the operating system must grant an
 available ephemeral port.
 
 The eventual host and port combination must uniquely identify the
--- a/host-port.js
+++ b/host-port.js
@@ -47,7 +47,7 @@ function validatePort(portNum, allowEmphemeral) {
         return 'Expected port to be a number';
     }
 
-    if (portNum < 0 || portNum > 65536) {
+    if (!(portNum >= 0 && portNum < 65536)) {
         return 'Expected port to be between 0 & 65536';
     }
 
--- a/host-port.js
+++ b/host-port.js
@@ -1,0 +1,105 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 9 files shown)


Raynos:
@wolski @CorgiMan 

I just updated this PR with documentation on what a valid hostPort is. I hope this helps make it more clear what ringpop should or should not do.

I think the biggest change here is not allowing you to use `localhost` in any tests, anywhere as can be seen by the refactoring of my own tests.

docs/channel.md
```diff
@@ -93,6 +93,9 @@ Starts listening on the given port and host.
 
 Both port and host are mandatory.
 
+Port must be a valid port as per [host-port rules](./host-port.md)
```
jcorbin: bullet list pls

host-port.js
```diff
@@ -0,0 +1,105 @@
...
+
+'use strict';
+
+var HostPort = {
+    validateHostPort: validateHostPort,
+    validateHost: validateHost,
+    validatePort: validatePort
+};
+
+module.exports = HostPort;
```
jcorbin: pls just mutate module.exports, not re-assign

host-port.js
```diff
@@ -0,0 +1,105 @@
...
+
+function validateHost(host, allowEmphemeral) {
+    if (typeof host !== 'string') {
+        return 'Expected host to be a string';
+    }
+
+    var hostParts = host.split('.');
```
jcorbin: there's no reason that this couldn't just be a `/\d+\.\d+\.\d+\.\d+/` regex, and then we wouldn't have to go to all the trouble of calling a function that re-strings each component.

host-port.js
```diff
@@ -0,0 +1,105 @@
...
+    if (hostParts.length !== 4) {
+        return 'Expected host to contain IPv4';
+    }
+
+    var i;
+    for (i = 0; i < 4; i++) {
+        if (!stringIsValidNumber(hostParts[i])) {
+            return 'Expected each IP section to be integer';
```
jcorbin: this error string is confusing, since an dotted-quad with non-numeric component isn't an IP address in the first place; calling it an "IP section" that's not an integer is confusing.

Please just use a simple regex, and say "Expected an IPv4 address" if not test

host-port.js
```diff
@@ -0,0 +1,105 @@
...
+
+    return null;
+}
+
+function validatePort(portNum, allowEmphemeral) {
+    if (typeof portNum !== 'number') {
+        return 'Expected port to be a number';
+    }
+
+    if (portNum < 0 || portNum > 65536) {
```
jcorbin: Please just `portNum > 0 && portNum < 65536`, and handle the zero case directly here.

host-port.js
```diff
@@ -0,0 +1,105 @@
...
+
+function validateHostPort(hostPort, allowEmphemeral) {
+    var reason;
+    if (typeof hostPort !== 'string') {
+        return 'Expected hostPort to be a string';
+    }
+
+    var parts = hostPort.split(':');
+    if (parts.length !== 2) {
+        return 'Expected hostPort to be {ip}:{port}';
```
jcorbin: please say "ipv4" throughout not just "ip"

host-port.js
Raynos: Rather just assign for consistency. mutation is non deterministic.

Raynos:
@jcorbin all the feedback addressed.

host-port.js
```diff
@@ -0,0 +1,95 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var validIPv4 = /\d+\.\d+\.\d+\.\d+/;
```
jcorbin: add `^...$` anchoring

jcorbin:
:ok:

rf:
lgtm

PR #236, opened 2015-12-05
Expose more init headers for debugging

Raynos:
Need these init headers to reason about what the other
side of the connection even is.

r: @jcorbin @rf @shannili @kriskowal


rf:
:+1:

PR #242, opened 2015-12-10
out_req: only mark end on last call response frame

Raynos:
This fixes a bug where we reported latency on the
first call response frame instead of the last call
response frame.

This was naughty.

r: @kriskowal @rf @jcorbin

### DIFF
--- a/out_request.js
+++ b/out_request.js
@@ -268,7 +268,7 @@ TChannelOutRequest.prototype.emitResponse = function emitResponse(res) {
         self.res.finishEvent.on(onFinished);
     }
 
-    self.responseEvent.emit(self, res);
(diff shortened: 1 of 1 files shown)


out_request.js
```diff
@@ -259,6 +259,25 @@ TChannelOutRequest.prototype.extendLogInfo = function extendLogInfo(info) {
...
+    self.res = res;
+    self.res.span = self.span;
+
+    if (!self.res.streamed) {
+        self.markEnd();
+    } else {
+        self.res.finishEvent.on(onFinished);
+    }
+
+    self.responseEvent.emit(self, res);
```
jcorbin: please `self.res` for consistency

PR #244, opened 2015-12-10
Fix a race conditions in operations.sanitySweep()

Raynos:
Because of nextTicks its possible for the cached `opKeys`
array and the `ops` object disagree about the state of the
world.

This is why I skip over any keys in `opKeys` that are not
own propertys of `ops`

r: @jcorbin @kriskowal @rf

### DIFF
--- a/operations.js
+++ b/operations.js
@@ -512,7 +512,7 @@ Operations.prototype._sweepOps = function _sweepOps(ops, direction, callback) {
                 direction: direction,
                 id: id
             });
-        } else if (op.timedOut) {
+        } else if (op && op.timedOut) {
             self.logger.warn('lingering timed-out operation', {
                 direction: direction,
                 id: id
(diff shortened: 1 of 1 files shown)


operations.js
```diff
@@ -498,6 +498,7 @@ Operations.prototype._sweepOps = function _sweepOps(ops, direction, callback) {
 
     nextOp(Object.keys(ops), 0, callback);
 
+    /*eslint complexity: 0*/
```
jcorbin: is this the same as `/* eslint-disable complexity */` ?

operations.js
```diff
@@ -506,7 +507,13 @@ Operations.prototype._sweepOps = function _sweepOps(ops, direction, callback) {
 
         var id = opKeys[i];
         var op = ops[id];
-        if (op === undefined) {
+
+        if (!Object.prototype.hasOwnProperty.call(ops, id)) {
+            setImmediate(deferNextOp);
```
jcorbin: no observability?

jcorbin:
:ok:

PR #245, opened 2015-12-10
WIP: as/json: ensure arg2 headers are map<string, string>

Raynos:
We have inconsistencies in our implementations.

Updating node to enforce that arg2 is not an arbitrary
JSON expression.

See change for python as well: https://github.com/uber/tchannel-python/pull/339

r: @kriskowal @rf @abhinav @breerly

### DIFF
--- a/as/json.js
+++ b/as/json.js
@@ -242,10 +242,31 @@ TChannelJSON.prototype.register = function register(
     }
 };
 
+function stringifyArg2(headers) {
+    headers = headers || {};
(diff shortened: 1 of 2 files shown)


abhinav:
lgtm

Raynos:
I broke the dependencies :(

aka: tcurl, hyperbahn.

I can do a dance to fix them.

Raynos:
RIP

PR #246, opened 2015-12-10
add TChannelThrift.parseException()

Raynos:
Enabled application developers to hook into our shouldAppRetry
logic when doing thrift encoded stuff.

It's kind of hard to actually expose it "optimally" but at
least exposing a `parseException()` method allows people to
get the job done.

r: @jcorbin @kriskowal @malandrew

### DIFF
--- a/test/as-thrift.js
+++ b/test/as-thrift.js
@@ -244,33 +244,6 @@ allocCluster.test('send and receive a typed notOk', {
     });
 });
 
-allocCluster.test('send and receive a typed notOk', {
-    numPeers: 2
-}, function t(cluster, assert) {
-    var tchannelAsThrift = makeTChannelThriftServer(cluster, {
-        notOkTypedResponse: true
-    });
-
-    var client = cluster.channels[1].subChannels.server;
-
-    tchannelAsThrift.send(client.request({
-        serviceName: 'server',
-        hasNoParent: true
-    }), 'Chamber::echo', null, {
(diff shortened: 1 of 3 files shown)


malandrew:
LGTM

test/as-thrift.js
```diff
@@ -130,6 +130,93 @@ allocCluster.test('send and receive a notOk', {
...
+        assert.equal(res.body, 10);
+        assert.equal(res.headers.as, 'thrift');
+
+        assert.end();
+    });
+
+    function shouldRetry(req, res, info, retry, done) {
+        assert.equal(info.exception.value, 10);
+        assert.equal(info.exception.message, 'No echo');
+        assert.equal(info.typeName, 'noEcho');
```
kriskowal: We have an opportunity to make this not weird, I think. Maybe accept (res) return a bool? "res" being the result provided for onResponse at the Thrift layer?

Raynos:
```
function (req, res, info) {
  return err; // do not retry AND fail with this new err
  return false; // do not retry
  return true; // retry
}
```

What about this ^

Raynos: Spoke with kris.

```js
function (resp/*, rawReq, rawRes*/) {
  return false; // do not retry
  return true; // retry
}
```

Raynos: Addressed.

PR #247, opened 2015-12-11
Classify and downgrade bufrw connection errors

jcorbin:
r @kriskowal @raynos

### DIFF
--- a/errors.js
+++ b/errors.js
@@ -22,6 +22,7 @@
 
 var TypedError = require('error/typed');
 var WrappedError = require('error/wrapped');
+var bufrwErrors = require('bufrw/errors');
 
 var Errors = module.exports;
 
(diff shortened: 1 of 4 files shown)


errors.js
```diff
@@ -648,6 +648,10 @@ Errors.classify = function classify(err) {
         return err.codeName;
     }
 
+    if (err.type.indexOf('bufrw.') > -1) {
+        return 'NetworkError';
```
rf: Not `ProtocolError`?

jcorbin: Yeah, in particular what we're seeing is a bufrw short error in connection closing.  Seems like remotes side is closing mid-frame-write.

jcorbin:
Updated based on bufrw 0.9.27 to clarify read vs write vs internal errors

Raynos:
lgtm.

kriskowal:
so say we all

PR #252, opened 2015-12-12
handle case where there simply is no conn

Raynos:
I've seen this uncaught in prod where `conn` is undefined.

Not really sure what to do here. for now its a logger.warn()
and No peer available.

r: @jcorbin @kriskowal

### DIFF
--- a/lazy_relay.js
+++ b/lazy_relay.js
@@ -213,6 +213,12 @@ function onIdentified(err) {
     }
 
     var conn = self.peer.getInConnection(true);
+    if (!conn) {
+        self.logger.warn('onIdentified called on non-existing connection', self.extendLogInfo({}));
+        self.onError(errors.NoPeerAvailable());
+        return;
+    }
+
     if (!conn.remoteName) {
         // we get the problem
         self.logger.warn('onIdentified called on unidentified connection', self.extendLogInfo({}));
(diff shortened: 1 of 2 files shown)


lazy_relay.js
```diff
@@ -213,11 +213,16 @@ function onIdentified(err) {
     }
 
     var conn = self.peer.getInConnection(true);
+    if (!conn) {
+        self.logger.warn('onIdentified called on non-existing connection', self.extendLogInfo({}));
```
jcorbin: - use `peer.extendLogInfo` too
- also it's regrettable that `PeerDrain#extendLogInfo` isn't called form `Peer#extendLogInfo`, I'd suggest fixing that since I would suspect maybe draining is to blame for this?

lazy_relay.js
```diff
@@ -213,11 +213,16 @@ function onIdentified(err) {
...
+        self.onError(errors.NoPeerAvailable());
+        return;
+    }
+
     if (!conn.remoteName) {
         // we get the problem
         self.logger.warn('onIdentified called on unidentified connection', self.extendLogInfo({}));
-    }
-    if (conn.closing) {
+    } else if (conn.closing) {
```
jcorbin: why this... also same remark, probably both of these warn log sites could benefit from `Peer#extendLogInfo`

jcorbin:
:ok: but we probably do want the peer / drain info

jcorbin:
Looks like this fixes #251

Raynos:
Addressed feedback.

PR #254, opened 2015-12-14
Expose getServiceEndpoints from tchannelAsThrift

### LINKED ISSUE #249: Implement tchannelAsThrift.getEndpointNames()
@kriskowal 

TChannel 3.5.19 now depends on thriftrw 3.x which is a breaking change. As such, anybody consuming the output of `TChannelAsThrift` will break. I believe this should justify a major version bump.

Willyham:
@kriskowal Fixes #249 

Depends on https://github.com/uber/thriftrw-node/pull/101

### DIFF
--- a/as/thrift.js
+++ b/as/thrift.js
@@ -353,6 +353,10 @@ function send(request, endpoint, outHead, outBody, callback) {
     }
 };
 
+TChannelAsThrift.prototype.getServiceEndpoints = function getEndpoints() {
+    return this.spec.getServiceEndpoints();
+};
+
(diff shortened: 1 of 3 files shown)


Raynos:
lgtm.

kriskowal:
We will need to roll the implementation of getServiceEndpoints into ThriftRW or implement it here instead.

kriskowal:
Looked deeper. Thanks—we will need to roll a release and uprev this branch to land.

kriskowal:
I’ve pushed a commit to uprev thriftrw. Standing by.

PR #255, opened 2015-12-14
Drop new-less support from many internal classes

jcorbin:
- ArgStream
- BatchStatsd
- InArgStream
- OutArgStream
- StreamArg
- TChannel(Self)Peer
- TChannelSelfConnection
- TChannelV2Handler
- TChannel{Root,Sub}Peers
- v2/ArgsRW
- v2/HeaderRW
- all tracing classes in trace/{agent,span}.js

r @raynos @kriskowal

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -31,10 +31,6 @@ BatchStatsd.BaseStat = BaseStat;
 var LENGTH_ARRAYS = {};
 
 function BatchStatsd(options) {
-    if (!(this instanceof BatchStatsd)) {
-        return new BatchStatsd(options);
-    }
-
     var self = this;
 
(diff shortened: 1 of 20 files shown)


lib/statsd.js
```diff
@@ -31,10 +31,6 @@ BatchStatsd.BaseStat = BaseStat;
 var LENGTH_ARRAYS = {};
 
 function BatchStatsd(options) {
-    if (!(this instanceof BatchStatsd)) {
```
Raynos: Note: may break hyperbahn upstream; trivial enough to fix.

kriskowal:
indeed, looks safe

Raynos:
legit. looks safe.

PR #256, opened 2015-12-14
Bring v2/*.js into eslint compliance

jcorbin:
I excepted curly rather than yet more churn.

r @raynos @kriskowal

### DIFF
--- a/v2/checksum.js
+++ b/v2/checksum.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+/* eslint-disable curly */
+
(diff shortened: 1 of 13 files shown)


v2/checksum.js
```diff
@@ -20,6 +20,8 @@
 
 'use strict';
 
+/* eslint-disable curly */
```
Raynos: we really should either just fix the call sites or disable this globally.

Raynos:
lgtm.

PR #257, opened 2015-12-14
Drop use of `var self` from v2/*.js

jcorbin:
r @kriskowal @raynos


kriskowal:
:ok: I truly hope this does something magical on production graphs.

Raynos:
legit.

PR #258, opened 2015-12-14
Drop .self pattern from many constructors

jcorbin:
r @Raynos @kriskowal

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -40,56 +40,57 @@ function TChannelConnection(channel, socket, direction, socketRemoteAddr) {
         'refusing to create self connection'
     );
 
-    var self = this;
-    TChannelConnectionBase.call(self, channel, direction, socketRemoteAddr);
-    self.identifiedEvent = self.defineEvent('identified');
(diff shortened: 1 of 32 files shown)


Raynos:
Yep. #legit. story checks out.

connection.js
```diff
@@ -40,56 +40,57 @@ function TChannelConnection(channel, socket, direction, socketRemoteAddr) {
         'refusing to create self connection'
     );
 
-    var self = this;
```
kriskowal: This occurrence appears to still be needed for callbacks below.

connection.js
```diff
@@ -40,56 +40,57 @@ function TChannelConnection(channel, socket, direction, socketRemoteAddr) {
...
     }
 
-    self.socket = socket;
-    self.ephemeral = false;
-    self.initHeaders = null;
+    this.socket = socket;
+    this.ephemeral = false;
+    this.initHeaders = null;
 
+    var self = this;
```
kriskowal: Found it. nm.

kriskowal:
roger :+1: ebert :+1:

PR #262, opened 2015-12-21
Goodbye self connection

Raynos:
This PR finally gets rid of SelfConnection and SelfPeer

r: @jcorbin @kriskowal


jcorbin:
:hammer:

test/send.js
```diff
@@ -313,39 +313,53 @@ allocCluster.test('request().send() to self', 1, function t(cluster, assert) {
...
-            resHead: 'head2', resBody: 'msg2',
-            resOk: false,
-            opts: {
-                host: one.hostPort,
-                serviceName: 'one'
+    subOne.waitForIdentified({
+        host: one.hostPort
+    }, onIdentified);
+
+    function onIdentified(err) {
```
kriskowal: unfortunate

kriskowal:
:ok: Are you sure about changing the Bad Request strings?

Raynos:
@kriskowal the bad request strings were regressed in a commit that joshua did that broke tcurl.

PR #263, opened 2015-12-23
Remove overzealous err is Error check from AsThrift

kriskowal:
r @Raynos @jcorbin


Raynos:
lgtm.

Raynos:
@jcorbin I suspect ^ this is a regression because we "broke" error frames by them not being errors anymore.

Raynos:
Also I am sorry. Having an `isError()` gaurd was stupid.

PR #265, opened 2015-12-25
add an inspect method to channel

Raynos:
I accidentally evaluated `tchannel` in a REPL.

It spend 60 seconds printing the 200MB object with all
the peers & connections.

I added an inspect method to avoid accidentally evaluating
really really expensive objects.

r: @jcorbin @kriskowal


rf:
:+1:

jcorbin:
:ok: how about adding a `#toString` while you're at it?

PR #269, opened 2016-01-20
Circuit breaking cleanup

Raynos:
This PR removes some circuit breaking from tchannel that
is now upstream in hyperbahn.

See hyperbahn PR: https://github.com/uber/hyperbahn/pull/215

r: @kriskowal @rf


jcorbin:
:ok: if you say you've tested the circular test fail outside travis

PR #270, opened 2016-01-20
Add more useful information to host port validation

jcorbin:
r @raynos


host-port.js
```diff
@@ -61,29 +61,29 @@ function validatePort(portNum, allowEmphemeral) {
 function validateHostPort(hostPort, allowEmphemeral) {
     var reason;
     if (typeof hostPort !== 'string') {
-        return 'Expected hostPort to be a string';
+        return 'Expected hostPort to be a string, got ' + typeof hostPort + ' instead';
```
Raynos: i think we want to print the value rather then the type.

Raynos:
lgtm.

PR #273, opened 2016-01-22
update dependencies for node4 compat

Raynos:
This will get tchannel-node to work on node@4.2

r: @jcorbin @rf @matt-esch


rf:
:+1: looks gr8

PR #274, opened 2016-01-25
perf: use array instead of typed array

Raynos:
It turns out that typed arrays are not optimized.
This makes the code faster.

r: @jcorbin @rf


rf:
:+1:

PR #271, opened 2016-01-21
add methods to v2/lazy_frame that do caching

Raynos:
Most of these changes are straight forward and
should improve perf.

The change to the header parsing hot code might
reduce perf.. not really sure, need more full
benchmarks.

I tried to microbench my change to header parsing
to make sure the new method still get's inlined.

I got:

```
raynos at raynos-ThinkPad-T440p  ~/uber/tchannel-node on lazy-caching*
$ cat inline | grep 'findOffset called'
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Inlined findOffset called from getValue.
Did not inline findOffset called from getValue (target not inlineable).
```

Not sure why it bailed the inline out...

r: @jcorbin @rf

PR #275, opened 2016-01-26
perf: Implement only readServiceStr()

Raynos:
This is a subset of https://github.com/uber/tchannel-node/pull/271

This only implements `readServiceStr()`; I've added a benchmark
for `readServiceStr()`.

It looks like the new implementation is about 2x faster, mostly
because:

 - no allocations of `ReadResult()`
 - using `utf8Slice()` directly.

I think @rf might be working on porting this fixes back
to `bufrw` itself; this will be great.

r: @jcorbin @rf

### DIFF
--- a/v2/call.js
+++ b/v2/call.js
@@ -127,25 +127,30 @@ CallRequest.RW.lazy.readServiceStr = function lazyReadServiceStr(frame) {
         return frame.cache.serviceStr;
     }
 
-    if (frame.size < CallRequest.RW.lazy.serviceOffset + 1) {
-        return null;
+    var serviceStrEnd = null;
+    if (frame.cache.headerStartOffset) {
+        serviceStrEnd = frame.cache.headerStartOffset;
+    } else {
+        if (frame.size < CallRequest.RW.lazy.serviceOffset + 1) {
+            return null;
+        }
+        var strLength = frame.buffer.readUInt8(
+            CallRequest.RW.lazy.serviceOffset, false
+        );
+        serviceStrEnd = CallRequest.RW.lazy.serviceOffset + 1 + strLength;
+        frame.cache.headerStartOffset = serviceStrEnd;
     }
-    var strLength = frame.buffer.readUInt8(
-        CallRequest.RW.lazy.serviceOffset, false
-    );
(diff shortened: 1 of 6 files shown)


v2/call.js
```diff
@@ -100,19 +122,48 @@ CallRequest.RW.lazy.readService = function lazyReadService(frame) {
...
+        return frame.cache.serviceStr;
+    }
+
+    if (frame.size < CallRequest.RW.lazy.serviceOffset + 1) {
+        return null;
+    }
+    var strLength = frame.buffer.readUInt8(
+        CallRequest.RW.lazy.serviceOffset, false
+    );
+    var end = CallRequest.RW.lazy.serviceOffset + 1 + strLength;
```
jcorbin: use the cache value if set rather than reading it

v2/call.js
```diff
@@ -100,19 +122,48 @@ CallRequest.RW.lazy.readService = function lazyReadService(frame) {
...
-        return res;
+    if (frame.cache.headerStartOffset !== null) {
+        offset = frame.cache.headerStartOffset;
+    } else {
+        // SKIP service~1
+        var res = bufrw.str1.sizerw.readFrom(frame.buffer, offset);
+        if (res.err) {
+            return res;
+        }
+        offset = res.offset + res.value;
```
jcorbin: cache the value for the other path

Raynos:
Addressed review feedback.

v2/call.js
```diff
@@ -100,19 +122,54 @@ CallRequest.RW.lazy.readService = function lazyReadService(frame) {
...
-
-    // SKIP service~1
-    var res = bufrw.str1.sizerw.readFrom(frame.buffer, offset);
-    if (res.err) {
-        return res;
+    if (frame.cache.headerStartOffset !== null) {
+        offset = frame.cache.headerStartOffset;
+    } else {
+        // SKIP service~1
+        var res = bufrw.str1.sizerw.readFrom(frame.buffer, offset);
```
rf: Can we avoid the `res` object alloc here? Why not just a standard `Buffer#readUint8` like above?

Raynos: This is the old path; I'm not updating it in this PR. Just adding offset caching.

PR #276, opened 2016-01-26
Add thrift/discover query to Hyperbahn client

dansimau:
Initial working version of discover for the Hyperbahn client.

Note: there are no tests. I wasn't sure how to test this. We should probably test this. I have only verified the success path is working.


Raynos:
@dansimau 

The only way to test this is to `npm link tchannel` in hyperbahn and contribute to https://github.com/uber/hyperbahn/tree/master/test/hyperbahn-client

It's a bit messy, we never figured out how to migrate `hyperbahn-client` somewhere with tests.

hyperbahn-client.js
```diff
@@ -445,6 +458,52 @@ function advertiseAgain(delay) {
...
+    var req = self.newRequest(opts);
+
+    self.tchannelThrift.send(req, 'Hyperbahn::discover', null, {
+        query: {
+            serviceName: self.serviceName
+        }
+    }, discoverInternalCb);
+
+    function discoverInternalCb(err, res) {
+        self.logger.info('in discoverInternalCb');
```
Raynos: I don't think this info() is needed outside of testing.

hyperbahn-client.js
```diff
@@ -148,6 +154,7 @@ function HyperbahnClient(options) {
             preferConnectionDirection: 'in'
         });
     self.tchannelJSON = TChannelJSON();
+    self.tchannelThrift = TChannelThrift({source: thriftSource});
```
Raynos: nitpick: whitespace conventions are:

```
self.tchannelThrift = TChannelThrift({
    source: thriftSource
});
```

Raynos:
This seems sensible. We do want a test, but it has to be upstream in hyperbahn...

dansimau:
Tests add here:
https://github.com/uber/hyperbahn/pull/223

dansimau:
Hey @Raynos, what are the next steps here to get this merged?

hyperbahn/utils.js
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
Raynos: A utils.js file is an anti-pattern; I'm just going to push a commit to inline this function.

Raynos:
Getting this past travis is proving difficult :) It's good to go though.

PR #277, opened 2016-01-26
Perf: Optimized readCallerNameStr() for lazy path

Raynos:
This is a faster version of `readCallerNameStr()` for the lazy
path.

It's about 3x faster then the existing code.

 - Avoid creating a slice() and use toString() directly
 - Use faster toString() ( utf8Slice )
 - Avoid creating Result objects.

r: @rf @jcorbin


jcorbin:
lgtm, tidy up to taste

PR #278, opened 2016-01-27
add cached readArg1Str()

Raynos:
This is the third & final method to allow reading of
arg1 efficiently.

Going to add benchmarks whilst waiting for reviews.

r: @jcorbin @rf

### DIFF
--- a/v2/call.js
+++ b/v2/call.js
@@ -318,8 +318,8 @@ CallRequest.RW.lazy.readArg1Str = function readArg1Str(frame) {
     var csumType = frame.buffer.readUInt8(offset, false);
     offset += 1;
 
-    if (csumType !== 0x00) {
-        offset += 4;
+    if (csumType !== Checksum.Types.None) {
+        offset += Checksum.offsetWidth(csumType);
     }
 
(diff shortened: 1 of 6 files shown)


rf:
lgtm

Raynos:
Ran the benchmarks.

readArg1Str() is about 4x faster.

All 4 methods together is about 3x faster.

v2/call.js
```diff
@@ -298,6 +298,48 @@ function readUInt8String(frame, offset) {
...
+
+    var offset = frame.cache.csumStartOffset;
+
+    if (frame.size < offset + 1) {
+        return null;
+    }
+    var csumType = frame.buffer.readUInt8(offset, false);
+    offset += 1;
+
+    if (csumType !== 0x00) {
```
jcorbin: please use the constant

v2/call.js
```diff
@@ -298,6 +298,48 @@ function readUInt8String(frame, offset) {
...
+        offset += 4;
```
jcorbin: please use checksum utility function or simple lookup table by type for this, don't hardcode

PR #279, opened 2016-01-27
Finally use the faster versions

Raynos:
Faster tracing is 3x faster.

 - Use the faster versions
 - Also faster ttl
 - Also faster tracing

r: @jcorbin @rf


rf:
lgtm

PR #280, opened 2016-01-28
remember to inherit ArgRW from BufferRW

rf:
This is required for TChannel to work with the zero-alloc bufrw.

r @Raynos @jcorbin


Raynos:
lgtm.

PR #281, opened 2016-01-28
Improve req/res log augmentation

jcorbin:
To get more data on an edge case @uber/ringpop team is seeing.

cc @raynos

### DIFF
--- a/v2/handler.js
+++ b/v2/handler.js
@@ -103,6 +103,24 @@ function TChannelV2Handler(options) {
 
 util.inherits(TChannelV2Handler, EventEmitter);
 
+TChannelV2Handler.prototype.justExtendLogInfo = function extendLogInfo(info) {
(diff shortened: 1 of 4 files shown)


v2/handler.js
```diff
@@ -705,12 +708,8 @@ function validateCallResponseFrame(res) {
         assert(res.headers && res.headers.as,
             'Expected the "as" transport header to be set for response');
     } else if (!res.headers || !res.headers.as) {
-        this.logger.error('Expected "as" header to be set for response', {
-            code: res.code,
-            remoteName: this.remoteName,
-            arg1: this.inreq.endpoint,
```
jcorbin: btw this was a subtle uncaught exception waiting to happen in a rare error handling path

Raynos:
lgtm; ship it.

PR #282, opened 2016-01-28
WIP: v2/handler: pass flush callback through pushFrame

jcorbin:
For debugging by @thanodnl

cc @raynos


Raynos:
I remember getting rid of this cb because it wasn't used.

I suspect it's not needed and the pending write queue check for >100 pending writes solves any problems of "did the write cb never get called".

PR #283, opened 2016-01-29
v2/lazy_frame: avoid out of bounds reads

Raynos:
We have gotten trolled multiple times by this
uncaught.

r: @rf @jcorbin


rf:
cool sure :+1:

PR #284, opened 2016-01-29
trace: do a proper bit flag check

Raynos:
Previously we assumed that there was only ever one flag.
This is not future proof.

r: @jcorbin @rf


rf:
sure :+1:

PR #286, opened 2016-02-02
lazy_relay: emit inbound stats for errorFrame path

Raynos:
Previously we only emitted inbound stats for requests that
we forward or error frames that we forward.

Now we do observability earlier and in more places to ensure
that we observe all inbound frames and all outbound error
frames.

r: @jcorbin @rf


jcorbin:
:+1:

PR #287, opened 2016-02-03
lazy_relay: only emit latency after flushing to socket

Raynos:
I made an oops. We need to write to the connection
and only then measure latency.

r: @jcorbin @rf


Raynos:
RIP

PR #288, opened 2016-02-03
connection: emit stats when sending error frame

Raynos:
Another piece of work towards complete visibility
around error frames.

r: @jcorbin @rf


jcorbin:
lgtm :+1:

PR #291, opened 2016-02-16
Fix node 0.10.42 bug around Content-Length

Raynos:
node 0.10.42 started enforcing HTTP 1.1 more strictly
and says that a request must NOT have both a
Content-Length and a Chunked-Encoding header.

This caused an issue for us in the `as/http`
implementation since we forward incoming requests
to lb_pool. When we forward we take a the 
Chunked-Encoding header and pass it to lb_pool as well
as passing it a Buffer for the body.

lb_pool will set the Content-Length header and this
creates an issue for the HTTP parser on the other
end.

In reality the `forwardToLBPool()` method does buffering
of the incoming HTTP request and therefore must strip
the Chunked-Encoding header as it's no longer doing
Chunked-Encoding

r: @rf @jcorbin @kriskowal

### DIFF
--- a/as/http.js
+++ b/as/http.js
@@ -448,6 +448,11 @@ TChannelHTTP.prototype._forwardToLBPool = function _forwardToLBPool(options, inr
     }
     options.encoding = null;
     var data = inreq.bodyStream || inreq.bodyArg; // lb_pool likes polymorphism
+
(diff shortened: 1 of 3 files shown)


jcorbin:
lgtm

PR #293, opened 2016-02-19
WIP: Associate ownership to error frames.

Raynos:
This changes marks the owner of an error frame inside
the message string.

The approach taken is to adjust the low level error
frame serialization to make ownerName mandatory.

From there on up we find all places where error frames
are send and verify whom the owner should be.

More work needs to be done on this branch.


Raynos:
RIP

PR #294, opened 2016-02-20
Log willSample replacer

rf:
This adds
- `replacer.js`, a script which replaces log calls with `if (logger.willSample('level', 'blah')) logger.level('blah')`
- new travis mode for running test suite with log calls rewritten
- converts loggers passed to TChannel or hyperbahn-client to Larch interface compatible loggers using Larch's interface converter

To test the willSample replacement, clone this branch and run `replacer.sh`.  You'll see a bunch of changes like

``` javascript
     function onError(err) {
         sent = true;
-        self.logger.warn('Handling request failed', {
+        if (self.logger.willSample('warn', 'Handling request failed')) self.logger.swarn('Handling request failed', {
             error: err
         });
```

Note that the link tests are going to fail since the Larch interface converter doesn't recognize `DebugLogtron`s that don't follow the `Logtron` interface, i.e. are pre-5.2.0, which includes https://github.com/Raynos/debug-logtron/pull/4. TCurl and Hyperbahn still have an old version of `DebugLogtron`.

r @Raynos @jcorbin @kriskowal

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "test": "npm run check-licence && npm run lint -s && npm run cover -s && npm run check-benchmark -s",
     "benchmark": "echo '!!! DEPRECATED: Better to just run `node benchmarks` directly' >&2; node benchmarks",
     "hyperbahn-link-test": "./test/link_hyperbahn.sh",
+    "log-replacer-test": "./replacer.sh; node test",
     "tcurl-link-test": "./test/link_tcurl.sh",
     "check-benchmark": "node benchmarks -- -r 10000 -p 100",
     "take-benchmark": "make -C benchmarks take",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -64,6 +66,7 @@
     "lb_pool": "^1.4.0",
     "ldjson-stream": "^1.2.1",
     "loadtest": "1.2.14",
+    "logtron": "^8.11.1",
```
Raynos: Why?

Raynos:
This seems reasonable.

package.json
rf: Larch's logtron backend requires `logtron/entry`. This is probably wrong. Maybe I should just copy that class def into Larch.

Raynos: please copy lal.

PR #295, opened 2016-02-20
fix race conditions in tchannel and hyperbahn

rajeshsegu:
We saw exception in case where the 'listening' was triggered after the tchannel is closed. Here is the fix for the race condition.


Raynos:
lgtm.

Raynos:
2.10.8

PR #296, opened 2016-02-21
`LazyRelay{In,Out}Req` Pooling

rf:
Inspired by Josh's PR, I tried to take a minimal stab at object pooling. This just pools `LazyRelay{In,Out}Req` objects. My pooling implementation also submits stats for all pools.

~~There's still some misplaced frees right now causing test failures, so I need to slog on this a little more to get it right.~~

This now passes all tests.

r @Raynos @jcorbin @kriskowal

### DIFF
--- a/lazy_relay.js
+++ b/lazy_relay.js
@@ -23,6 +23,7 @@
 var errors = require('./errors');
 var v2 = require('./v2');
 var stat = require('./stat-tags.js');
+var ObjectPool = require('./lib/object_pool');
 var ReadResult = require('bufrw').ReadResult;
 var WriteResult = require('bufrw').WriteResult;
 
@@ -40,6 +41,46 @@ module.exports = {
 
 /*eslint max-statements: [2, 40]*/
 function LazyRelayInReq(conn, reqFrame) {
+    this.channel = null;
+    this.conn = null;
+    this.start = null;
+    this.remoteAddr = null;
+    this.logger = null;
+    this.peer = null;
+    this.outreq = null;
+    this.reqFrame = null;
+    this.id = null;
+    this.serviceName = null;
+    this.callerName = null;
+    this.timeout = null;
+    this.alive = null;
+    this.operations = null;
+    this.timeHeapHandle = null;
+    this.endpoint = null;
+    this.error = null;
+    this.tracing = null;
+    this.reqContFrames = [];
+    this.hasRead = null;
+
+    this.boundExtendLogInfo = extendLogInfo;
+    this.boundOnIdentified = onIdentified;
+
+    var self = this;
+
+    function extendLogInfo(info) {
+        return self.extendLogInfo(info);
+    }
+
+    function onIdentified(err) {
+        if (err) {
+            self.onError(err);
+        } else {
+            self.onIdentified();
+        }
+    }
+}
+
+LazyRelayInReq.prototype.reset = function reset(conn, reqFrame) {
     this.channel = conn.channel;
     this.conn = conn;
     this.start = conn.timers.now();
@@ -58,27 +99,40 @@ function LazyRelayInReq(conn, reqFrame) {
     this.endpoint = '';
     this.error = null;
     this.tracing = null;
-    this.reqContFrames = [];
+    this.reqContFrames.length = 0;
     this.hasRead = false;
-
-    this.boundExtendLogInfo = extendLogInfo;
-    this.boundOnIdentified = onIdentified;
     this.circuit = reqFrame.circuit;
+};
 
-    var self = this;
-
-    function extendLogInfo(info) {
-        return self.extendLogInfo(info);
+LazyRelayInReq.prototype.clear = function clear() {
+    if (this.outreq && !this.outreq._objectPoolIsFreed) {
+        this.outreq.free();
     }
 
-    function onIdentified(err) {
-        if (err) {
-            self.onError(err);
-        } else {
-            self.onIdentified();
-        }
-    }
-}
+    this.channel = null;
+    this.conn = null;
+    this.start = null;
+    this.remoteAddr = null;
+    this.logger = null;
+    this.peer = null;
+    this.outreq = null;
+    this.reqFrame = null;
+    this.id = null;
+    this.serviceName = null;
+    this.callerName = null;
+    this.timeout = null;
+    this.alive = null;
+    this.operations = null;
+    this.timeHeapHandle = null;
+    this.endpoint = null;
+    this.error = null;
+    this.tracing = null;
+    this.reqContFrames.length = 0;
+    this.hasRead = null;
+    this.circuit = null;
+};
+
+ObjectPool.setup({Type: LazyRelayInReq, maxSize: 200});
 
 LazyRelayInReq.prototype.type = 'tchannel.lazy.incoming-request';
 
@@ -264,7 +318,8 @@ LazyRelayInReq.prototype.forwardTo =
 function forwardTo(conn) {
     var self = this;
 
-    self.outreq = new LazyRelayOutReq(conn, self);
+    self.outreq = LazyRelayOutReq.alloc();
+    self.outreq.reset(conn, self);
 
     var ttl = self.updateTTL(self.outreq.start);
     if (!ttl || ttl < 0) {
@@ -344,8 +399,10 @@ function onReadError(err) {
         );
     }
 
+    var conn = self.conn;
     self.onError(err);
-    self.conn.resetAll(err);
+
+    conn.resetAll(err);
 };
 
 LazyRelayInReq.prototype.onError =
@@ -356,6 +413,7 @@ function onError(err) {
         self.logger.warn('dropping error from dead relay request', self.extendLogInfo({
             error: err
         }));
+        self.free();
         return;
     }
 
@@ -396,6 +454,7 @@ function handleFrameLazily(frame) {
 
     if (!self.alive) {
         self.logger.warn('dropping frame from dead relay request', self.extendLogInfo({}));
+        self.free();
         return;
     }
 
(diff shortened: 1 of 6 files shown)


lib/object_pool.js
```diff
@@ -0,0 +1,170 @@
...
+ObjectPool.pools = [];
+ObjectPool.timer = null;
+ObjectPool.bootstrapped = false;
+ObjectPool.refs = 0;
+
+ObjectPool.setup = function setup(options) {
+    var pool = new ObjectPool(options);
+    options.Type.alloc = alloc;
+
+    options.Type.prototype.free = function freeThisObj() {
```
kriskowal: Perhaps consider using just one style of closure, like alloc or like free[ThisObj[ect]].

kriskowal:
:eyes: tests. be sure to squash the wip. ping me for another review.

lazy_relay.js
```diff
@@ -344,8 +399,10 @@ function onReadError(err) {
         );
     }
 
+    var conn = self.conn;
     self.onError(err);
-    self.conn.resetAll(err);
+
+    conn.resetAll(err);
```
jcorbin: why this change, looks like churn?

lazy_relay.js
```diff
@@ -644,6 +738,7 @@ function handleFrameLazily(frame) {
     frame.setId(self.inreq.id);
     self.inreq.conn.writeToSocket(frame.buffer);
     if (frame.bodyRW.lazy.isFrameTerminal(frame)) {
+        self.alive = false;
```
jcorbin: this looks like a prior-existing bug that we didn't mark dead here?

lazy_relay.js
```diff
@@ -523,6 +582,22 @@ function _observeCallReqContFrame(frame) {
...
+    this.remoteAddr = null;
+    this.logger = null;
+    this.inreq = null;
+    this.id = null;
+    this.serviceName = null;
+    this.callerName = null;
+    this.timeout = null;
+    this.operations = null;
+    this.timeHeapHandle = null;
+    this.alive = null;
```
jcorbin: is having alive be essentially a "tri-state" intentional? I think it seems useful, but just wanted to check

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+
+function ObjectPool(options) {
+    assert(typeof options === 'object', 'expected options object');
+
+    this.Type = options.Type;
+    assert(typeof this.Type === 'function', 'expected options.Type to be constructor function');
+    assert(typeof this.Type.prototype.reset === 'function', 'expected options.Type to have reset method');
+    assert(typeof this.Type.prototype.clear === 'function', 'expected options.Type to have clear method');
+
+    this.maxSize = options.maxSize || 1000;
```
jcorbin: factor out `DEFAULT_MAX_SIZE`

lib/object_pool.js
jcorbin: `freeThisObj` could be hoisted to module level fwiw

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+    ObjectPool.timers = options.timers;
+
+    ObjectPool.timer = ObjectPool.timers.setTimeout(
+        ObjectPool.reportStats,
+        ObjectPool.reportInterval
+    );
+};
+
+ObjectPool.unref = function unref() {
+    ObjectPool.refs = Math.min(0, ObjectPool.refs - 1);
```
jcorbin: didn't you mean `Math.max` here?

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+    if (ObjectPool.refs === 0) {
+        ObjectPool.timers.clearTimeout(ObjectPool.timer);
```
jcorbin: may be worth nulling-after-clear for clarity

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+ObjectPool.reportStats = function reportStats() {
+    // Iterate over pools
+    // report their current size
+    var i;
+    var pool;
+    for (i = 0; i < ObjectPool.pools.length; i++) {
+        pool = ObjectPool.pools[i];
+        if (ObjectPool.statsd) {
+            ObjectPool.statsd.gauge(
+                'object-pools.' + pool.name + '.free',
```
jcorbin: all these strings could be constant-by-pool, so maybe we should have the inner part of this loop factored out as say `ObjectPool#reportStats(statsd)`

lazy_relay.js
rf: Setting everything to `null` is more of a safety thing -- I want to induce reference errors when we accidentally use an object while it's still part of the free list

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+};
+
+ObjectPool.reportStats = function reportStats() {
+    // Iterate over pools
+    // report their current size
+    var i;
+    var pool;
+    for (i = 0; i < ObjectPool.pools.length; i++) {
+        pool = ObjectPool.pools[i];
+        if (ObjectPool.statsd) {
```
jcorbin: hoist the if out and skip the entire loop

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+
+            // This stat is how we infer the maxSize property
+            ObjectPool.statsd.gauge(
+                'object-pools.' + pool.name + '.outstanding',
+                pool.outstanding
+            );
+        }
+    }
+
+    ObjectPool.timer = ObjectPool.timers.setTimeout(
```
jcorbin: do we skip setting a timer too if we have no statsd?

lazy_relay.js
rf: We didn't have an alive var in `OutReq`s before; I'm not sure we needed it

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+        inst = new this.Type();
+        inst._objectPoolIsFreed = false;
+        return inst;
+    }
+};
+
+ObjectPool.prototype.free = function free(inst) {
+    assert(!inst._objectPoolIsFreed, 'object pool double free');
+    inst._objectPoolIsFreed = true;
```
jcorbin: this mutates hidden classes, right?

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+
+    inst.clear();
+    if (this.outstanding <= this.maxSize) {
```
jcorbin: I think I want a stat in the contra-case here to know how often I'm blowing my pool limits?

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    ObjectPool.timer = ObjectPool.timers.setTimeout(
+        ObjectPool.reportStats,
+        ObjectPool.reportInterval
+    );
+};
+
+ObjectPool.prototype.get = function get() {
```
jcorbin: it'd be nice to have hit/miss stats from this

lazy_relay.js
rf: `LazyRelayInReq#onError` may `free()`, so we need to save a ref to `conn` in order to call `resetAll`.

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+    var inst;
+    this.outstanding += 1;
```
jcorbin: isn't `outstanding` wholly redundant (the complement of) `freeList.length`?  Seems like we could just not have it, and the whole class of "can't ever happen" bugs around it de-syncing from `freeList.length`.

lib/object_pool.js
```diff
@@ -0,0 +1,171 @@
...
+    if (this.freeList.length) {
+        inst = this.freeList.pop();
+        assert(inst._objectPoolIsFreed, 'instance retreived from pool is free');
+        inst._objectPoolIsFreed = false;
+        return inst;
```
jcorbin: so it's on the caller to call `.reset` right?  And that's necessary because resets are contextual?

lib/object_pool.js
rf: It does, but since *all* objects that are of a class setup to be pooled will have this property, it shouldn't have an affect on optimization since the functions will all be compiled against the map for `LazyRelayInReq'`.

lib/object_pool.js
rf: You would see outstanding > maxSize, and free !> 0

lib/object_pool.js
rf: No, outstanding is *total* instances, including those that couldn't be pooled. It's maximum value infers the maxSize property.

lib/object_pool.js
rf: Yes, `reset` is used for it's less-common meaning, a literal *re-*set. Objects will be cleared when `free()`d and `reset` is used to re-initialize.

lib/object_pool.js
```diff
@@ -0,0 +1,169 @@
...
+        ObjectPool.timers.clearTimeout(ObjectPool.timer);
+        ObjectPool.timer = null;
+    }
+};
+
+ObjectPool.reportStats = function reportStats() {
+    // Iterate over pools, report their current size
+
+    var i;
+    for (i = 0; i < ObjectPool.pools.length; i++) {
```
jcorbin: what happened to checking for if statsd is a thing?

jcorbin:
lgtm, I'm okay with not having the hit/miss stats initially like this, but think we may want them.

Maybe we could eventually have a `pool.hits` and `pool.misses` number that gets incremented under `.get`, and then reported and reset to zero by `.reportStats` ?

lib/object_pool.js
rf: `ObjectPool` can't be bootstrapped without a `statsd` because of the assertions, so the timer will never be set and this function will never be called if there's no `statsd`. The check was actually unnecessary.

PR #297, opened 2016-02-23
RFC: Carry response remoteAddr on TChannel{JSON,Thrift}Response

jcorbin:
So I needed this while writing a hyperbahn test to log which node I happened to
advertise to/register with.

Personally I feel like the specialized responses should either carry the entire
wire response object, or some sort of more general "context" then just
headers... things missing still:
- span
- start time
- any outgoing req parts

cc @kriskowal @raynos

### DIFF
--- a/as/json.js
+++ b/as/json.js
@@ -151,6 +151,7 @@ function TChannelJSONResponse(response, parseResult) {
         self.body = errors.ReconstructedError(parseResult.body);
     }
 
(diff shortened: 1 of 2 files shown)


Raynos:
This is a great first pass.

Carrying the raw response is a decent start, but we probably want to move towards collapsing the objects / inverting the relationship.

The raw response can use fields for decoded arg2/arg3/ etc.

PR #299, opened 2016-02-26
add stopWaitingForIdentified and use in LazyRelayInreq

rf:
Adds a method called `stopWaitingForIdentified` on peers. Uses a descriptor table to keep track of what needs to be released.

Note that there is a `free` added to the end of `LazyRelayInReq#onError`. This `free` would cause test failures but for the `stopWaitingForIdentified` method and its use in `LazyRelayInReq#clear`, hence the lack of tests; this condition is already sufficiently stressed.

This `free` added to the bottom of `onError` is (at least one of) the missing `free`s that are causing the object pool to not reuse objects right now in prod.

r @jcorbin @Raynos

### DIFF
--- a/lazy_relay.js
+++ b/lazy_relay.js
@@ -74,6 +74,10 @@ function LazyRelayInReq(conn, reqFrame) {
     }
 
     function onIdentified(err) {
+        // The ident descriptor will be cleared out of the peer when the ident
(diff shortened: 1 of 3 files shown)


jcorbin:
:ok:

PR #300, opened 2016-02-26
ObjectPool: use batch stats

rf:
Use batch stats in object pool. We don't actually pass a statsd to TChannel in Hyperbahn, so we have to use the batch statsd.

r @jcorbin @Raynos

### DIFF
--- a/stat-tags.js
+++ b/stat-tags.js
@@ -58,7 +58,8 @@ module.exports = {
     ConnectionsErrorsTags: ConnectionsErrorsTags,
     ConnectionsClosedTags: ConnectionsClosedTags,
     RelayLatencyTags: RelayLatencyTags,
-    HTTPHanlderBuildLatencyTags: HTTPHanlderBuildLatencyTags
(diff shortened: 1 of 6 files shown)


jcorbin:
I like stats, and I like them even better in batches :+1:

stat-tags.js
```diff
@@ -663,3 +664,17 @@ HTTPHanlderBuildLatencyTags.prototype.toStatKey = function toStatKey(prefix) {
...
+    this.host = '';
+    this.cluster = '';
+    this.version = '';
+
+    this.poolName = poolName;
+    this.statType = statType;
+}
+
+ObjectPoolTags.prototype.toStatKey = function toStatKey(prefix) {
+    return prefix + '.' + this.poolName + '.' + this.statType;
```
jcorbin: y u no catch dat suffix

jcorbin: oh, is it testonly?

jcorbin:
I still like thees batch stats :100:

PR #301, opened 2016-02-27
Pool test verification & missing free

rf:
Added a verification step to the end of test cluster and a debug mode for the object pool which makes sure all of the pools have no outstanding instances.

After adding this test, found a missing free, so added that.

This work builds on both #299 and #300. 

r @jcorbin @Raynos


rf:
Closing to open on #300 instead.

PR #302, opened 2016-02-27
Free LazyRelayInReqs upon NoPeerAvailable; Object Pool verification

rf:
Adds a verification step when tearing down a test cluster that ensures that all objects allocated using pools have been freed.

Adds a missing `LazyRelayInReq#free` upon no peer available.

Builds on #300 because they both modify the ObjectPool singelton.

r @raynos @jcorbin

### DIFF
--- a/lib/object_pool.js
+++ b/lib/object_pool.js
@@ -53,6 +53,9 @@ function ObjectPool(options) {
         this.name,
         'outstanding'
     );
+
+    // only used in debug mode
+    this.outstandingList = [];
 }
 
 ObjectPool.channel = null;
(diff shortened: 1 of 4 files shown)


lib/object_pool.js
```diff
@@ -178,4 +196,13 @@ ObjectPool.prototype.free = function free(inst) {
         this.freeList.push(inst);
     }
     this.outstanding -= 1;
+
+    var i;
+    if (ObjectPool.debug) {
+        for (i = 0; i < this.outstandingList.length; i++) {
+            if (this.outstandingList[i] === inst) {
+                this.outstandingList.splice(i, 1);
+            }
```
jcorbin: can there be dupes in the list?

jcorbin:
lgtm

lib/object_pool.js
rf: No; objects are only added when they're allocated. An object would have to be freed twice, but we have an assertion guard against that. There could be a return after we do the splice. I'll add that in.

PR #304, opened 2016-03-03
Implement minConnections as a step towards goalConnections

Raynos:
I've written a test to verify load balance distribution
between client & server in p2p mode.

As can be seen, without minConnections the load distribution
is absolutely terrible and it improves with minConnections

r: @jcorbin @kriskowal @rf

### DIFF
--- a/peer_heap.js
+++ b/peer_heap.js
@@ -30,8 +30,10 @@ module.exports = PeerHeap;
 // peer list.
 var dfsStack = [0, 1, 2];
 
-function PeerHeap(random) {
+function PeerHeap(peers, random) {
     this.array = [];
+    this.peers = peers || null;
+    this.hasMinConnections = peers ? peers.hasMinConnections : false;
 
     this.random = random || Math.random;
     assert(typeof this.random === 'function', 'PeerHeap expected random fn');
@@ -72,17 +74,17 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
         return null;
     }
 
-    if (self.array.length === 1) {
-        return self.choose1(threshold, filter);
-    }
-
-    if (self.array.length === 2) {
-        return self.choose2(threshold, filter);
-    }
-
+    var isSecondary = false;
     var chosenPeer = null;
+    var secondaryPeer = null;
     var highestProbability = 0;
+    var secondaryProbability = 0;
     var firstScore = self.array[0].peer.getScore();
+
+    var notEnoughPeers = false;
+    if (self.hasMinConnections) {
+        notEnoughPeers = self.peers.currentConnectedPeers < self.peers.minConnections;
+    }
 
     // Pointers into dfsStack
     var stackBegin = 0;
@@ -90,8 +92,18 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
 
     if (firstScore > threshold && (!filter || filter(self.array[0].peer))) {
         // Don't check first peer if it looks good, check its children though
-        chosenPeer = self.array[0].peer;
-        highestProbability = firstScore;
+
+        var firstPeer = self.array[0].peer;
+        isSecondary = notEnoughPeers && firstPeer.isConnected('out');
+
+        if (isSecondary) {
+            secondaryPeer = firstPeer;
+            secondaryProbability = firstScore;
+        } else {
+            chosenPeer = firstPeer;
+            highestProbability = firstScore;
+        }
+
         // The array is seeded with 0, 1, 2 so we just have to advance the
         // stack pointers
         stackBegin = 1;
@@ -104,7 +116,9 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
 
         var el = self.array[i];
 
-        if (!filter && self.rangehis[i] <= self.maxRangeStart) {
+        if (self.rangehis[i] <= self.maxRangeStart &&
+            (!filter && !notEnoughPeers)
+        ) {
             // This range ends before the range with the largest start begins,
             // so it can't possibly be chosen over any of the ranges we've
             // seen. All ranges below this one have a smaller end, so this
@@ -113,6 +127,8 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
         }
 
         if (!filter || filter(el.peer)) {
+            isSecondary = notEnoughPeers && el.peer.isConnected('out');
+
             // INLINE of TChannelPeer#getScore
             var lo = self.rangelos[i];
             var hi = self.rangehis[i];
@@ -122,9 +138,18 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
             }
             var probability = lo + ((hi - lo) * rand);
 
-            if ((probability > highestProbability) && (probability > threshold)) {
-                highestProbability = probability;
-                chosenPeer = el.peer;
+            if (
+                (probability > threshold) &&
+                !isSecondary ? (probability > highestProbability) :
+                               (probability > secondaryProbability)
+            ) {
+                if (isSecondary) {
+                    secondaryProbability = probability;
+                    secondaryPeer = el.peer;
+                } else {
+                    highestProbability = probability;
+                    chosenPeer = el.peer;
+                }
             }
         }
 
@@ -140,7 +165,7 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
         }
     }
 
-    return chosenPeer;
+    return chosenPeer || secondaryPeer;
 };
 /*eslint-enable complexity */
 /*eslint-enable max-statements */
--- a/peer_heap.js
+++ b/peer_heap.js
@@ -104,13 +104,15 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
 
         var el = self.array[i];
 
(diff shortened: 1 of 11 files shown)


channel.js
```diff
@@ -122,7 +122,8 @@ function TChannel(options) {
     this.emitConnectionMetrics =
         typeof this.options.emitConnectionMetrics === 'boolean' ?
         this.options.emitConnectionMetrics : false;
-    this.choosePeerWithHeap = this.options.choosePeerWithHeap || false;
+    this.choosePeerWithHeap = typeof this.options.choosePeerWithHeap === 'boolean' ?
```
jcorbin: interesting change, separate?

sub_peers.js
```diff
@@ -27,13 +28,67 @@ var PeerHeap = require('./peer_heap.js');
...
+
+    this.currentConnectedPeers = 0;
+
+    this.boundOnOutConnectionIncrement = boundOnOutConnectionIncrement;
+    this.boundOnOutConnectionDecrement = boundOnOutConnectionDecrement;
+
+    function boundOnOutConnectionDecrement(_, peer) {
+        self.onOutConnectionDecrement(peer);
+    }
+    function boundOnOutConnectionIncrement(_, peer) {
```
jcorbin: seems maybe a missed opportunity to transmit a delta number in the event, and replace two events with one?

sub_peers.js
```diff
@@ -27,13 +28,67 @@ var PeerHeap = require('./peer_heap.js');
...
+
+    for (var i = 0; i < peers.length; i++) {
+        var p = peers[i];
+
+        if (p.countConnections('out') > 0) {
+            count++;
+        }
+    }
+
+    assert(count !== this.currentConnectedPeers,
```
jcorbin: why is this a hard-throwing assertion, rather than an error, or simple string returning function?

Raynos: this was purely used for spot testing. a hard assertion bails the test suite visibly.

I think i should remove this function as its outlived its use.

sub_peers.js
```diff
@@ -125,7 +197,16 @@ TChannelSubPeers.prototype.chooseLinearPeer = function chooseLinearPeer(req) {
     for (var i = 0; i < hosts.length; i++) {
         var hostPort = hosts[i];
         var peer = self._map[hostPort];
-        if (!req || !req.triedRemoteAddrs || !req.triedRemoteAddrs[hostPort]) {
+
+        var shouldSkip = req && req.triedRemoteAddrs && req.triedRemoteAddrs[hostPort];
+        if (this.hasMinConnections) {
+            var notEnoughPeers = this.currentConnectedPeers < this.minConnections;
```
jcorbin: So we'll prefer to initiate a new connection, and wait for its init handshake to complete, before we send the current rpc?  Can't we decouple that so that we don't defer sending the current rpc just to open some new connections?

sub_peers.js
```diff
@@ -159,7 +240,9 @@ TChannelSubPeers.prototype.chooseHeapPeer = function chooseHeapPeer(req) {
     var self = this;
 
     var peer;
-    if (req && req.triedRemoteAddrs) {
+    if ((req && req.triedRemoteAddrs) ||
+        (this.hasMinConnections && this.currentConnectedPeers < this.minConnections)
```
rf: Seems like the second half of this or could be brought out and calculated once:
```
var notEnoughPeers = self.hasMinConnections && this.currentConnectedPeers < minConnections
```
Then you could use it in the filter function blow as well

peer_heap.js
```diff
@@ -104,13 +104,15 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
 
         var el = self.array[i];
 
-        if (self.rangehis[i] <= self.maxRangeStart) {
+        if (!filter && self.rangehis[i] <= self.maxRangeStart) {
```
rf: I need to think about it more but I'm not sure that this is what we want here. This actually removes any benefit from using the heap and devolves it into a filtered linear scan.

sub_peers.js
Raynos: We could but that increases complexity. 

If `minConnections=10` it will warm up pretty quick. The window of time where you wait for 1 init handshake or 10 init handshakes is similar.

I think its a micro op that we can do later.

peer_heap.js
Raynos: if in filter mode; yes. but we are not always in filter mode.

sub_peers.js
Raynos: We only go into _choosePeerSkipTried() if we are below the threshold so the vast majority of the time we go in the hotpath of `choose()`

I'll optimize it if it turns out to be a real problem.

channel.js
Raynos: https://github.com/uber/tchannel-node/pull/305

sub_peers.js
Raynos: good call. Will fix.

Raynos:
Here is a piece of motivation for this change:

We currently only rely on inbound connections from hyperbahn. This is great until you stop using hyperbahn. If you ever do peer to peer tchannel from A -> B then you get bad semantics because tchannel will re-use any open connections it has even if it just has one.

When tchannel has one connection the variance is huge, generally when 40 As are talking to 40 Bs only about 25/40 Bs will even have inbound connections because probability is really not in our favor. The load distribution is terrible, least pending selection wont work because there is only one peer. A lot of features of tchannel are just sub optimal.

This change fixes that my adding a mandatory `minConnections`. `minConnections` is treated similarly to `triedRemoteAddrs`. It's a skipping predicate when choosing a peer. `minConnections` in this context actually means `minConnectedPeers` and will skip a peer if it does not have at least `minConnections` number of connected peers. This means the first few requests for a channel will have some overhead by opening new connections but it will quickly stabilize when we have more connections then `minConnections` and we do normal peer selection as usual.

The changes to the peer heap are a little more involved. linear peer selection will look at all peers. The peer heap actually only looks at peers in a bucket (aka identified, connected, unconnected) and thus there might be less the `minConnections` peers per bucket (say only 1 or 2). This means the peer heap had to be changed so that when a filter is supplied we try all buckets until the filter becomes happy.

cc @rf

Raynos:
Another update on this branch, taking it in a different direction:

The `minConnectedPeers` logic shouldn't be a filter(). This can lead to a configuration where the `minConnectedPeers` is too high and we have less available peers. Instead the `minConnectedPeers` is now treated as a secondary set of peers for selection. 

We have favor the primary peers and choose one if available, however we will choose a secondary peer (i.e. a skipped peer that already has connections in this case) if available.

The second update that is made is that we still choose the peer with the highest score. We simply keep track of the "primary peer" and open a connection to it if the secondary peer score is higher. This effectively means we always choose the best peer but keep track of the next highest score peer we should connect to.

By connecting to it in the "background" we avoid a condition where there are more `minConnectedPeers` then healthy peers and we keep selecting unhealthy peers and get them stuck in an ECONNREFUSED loop.

r: @jcorbin @rf @kriskowal

Raynos:
Addded and ran benchmarks; no noticable difference.

channel.js
jcorbin: rebase on merge?

peer_heap.js
```diff
@@ -44,25 +46,6 @@ function PeerHeap(random) {
     this.maxRangeStart = 0;
 }
 
-PeerHeap.prototype.choose1 = function choose1(threshold, filter) {
```
jcorbin: goodbye micro op? we didn't need you anyways?

Raynos: i removed the micro op because the logic got messy.

I think optimizing for 1 or 2 peers isn't very common in prod.

peer_heap.js
```diff
@@ -104,13 +97,20 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
...
             // This range ends before the range with the largest start begins,
             // so it can't possibly be chosen over any of the ranges we've
             // seen. All ranges below this one have a smaller end, so this
             // range and any below it can't be chosen.
             continue;
-        } else if (!filter || filter(el.peer)) {
+        }
+
+        if (!filter || filter(el.peer)) {
+            isSecondary = notEnoughPeers && el.peer.isConnected('out');
```
rf: I did a lot of work to make sure this loop was only accessing small, packed arrays of numbers. Even reaching into an object like this will violate cache locality, and I'd really rather not have a function call that can't be inlined. Did you benchmark / flamegraph this?

peer_heap.js
```diff
@@ -120,9 +120,18 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
             }
             var probability = lo + ((hi - lo) * rand);
 
-            if ((probability > highestProbability) && (probability > threshold)) {
-                highestProbability = probability;
-                chosenPeer = el.peer;
+            if (
+                (probability > threshold) &&
+                !isSecondary ? (probability > highestProbability) :
+                               (probability > secondaryProbability)
```
rf: ternary *inside* if? plz don't

peer_heap.js
Raynos: yeah i benched it. very little impact. 

because `notEnoughPeers` is false the majority of the time.

Raynos:
Thanks for all the code review everybody :)

PR #305, opened 2016-03-04
channel: default choose peer with heap to true

Raynos:
This change makes the peer heap the default in the
tchannel library. This means all users of tchannel
will go fast by default.

r: @jcorbin @rf @kriskowal


rf:
:+1:

kriskowal:
:ok:

PR #306, opened 2016-03-04
Upgrade ThriftRW to 3.4.0

kriskowal:


kriskowal:
This ships a startup time improvement by removing the IDL parser generation from start up time.

Raynos:
yay.

PR #307, opened 2016-03-06
Shield againts calling `onError()` on null.

Raynos:
It might be possible for the same `inreq` to fail in multiple
ways in one tick of the event loop.

This means that other failures could free the inreq and it would
be gone.

Once we get to the nextTick we should do a best effort error
handling and if it's already freed assume that it failed with
a different but related error.

r: @jcorbin @rf


kriskowal:
:ok:

PR #308, opened 2016-03-10
Fix a bunch of uncaught exceptions.

Raynos:
There is a free() somewhere that isnt doing the right
thing. Freed objects are sticking around in pending
operations on the connection.

This causes a bunch of bugs with trying to error or
timeout a freed LazyOutReq/LazyInReq

I've added defense in depth to work around these
issues.

r: @jcorbin @rf


Raynos:
Merging because:

wat

kriskowal:
:ok:

PR #309, opened 2016-03-16
peer: add tryConnect() with backoff

Raynos:
There is an edge cases where the peer selection logic as
part of `minConnections` might try and open a lot of connections.

This can be quite expensive in terms of CPU & logging.

This change adds exponential backoff to trying to open a connection
on an unconnected peer when we have less connected peers then
minConnections.

There might be a scenario where you have minConnections: 4 and
numPeers: 5, except two of 5 peers are down. In that scenario
tchannel would try and open a lot of connections to the down
peers even though it should know that they are down.

The more general solution to this is remember whether a peer
is healthy or not.

r: @jcorbin @rf @kriskowal

### DIFF
--- a/peer.js
+++ b/peer.js
@@ -68,6 +68,9 @@ function TChannelPeer(channel, hostPort, options) {
     this.boundOnPendingChange = onPendingChange;
     this.scoreRange = null;
 
+    this.nextConnAttemptTime = 0;
+    this.nextConnAttemptDelay = 0;
+
     this.waitForIdentifiedListeners = [];
 
     this.reportInterval = options.reportInterval || DEFAULT_REPORT_INTERVAL;
@@ -408,6 +411,45 @@ function connect(outOnly) {
 TChannelPeer.prototype.connectTo = function connectTo() {
     var self = this;
     return self.connect(true);
+};
+
+TChannelPeer.prototype.tryConnect = function tryConnect() {
+    var self = this;
+
+    var rightNow = Date.now();
+    if (rightNow < self.nextConnAttemptTime) {
+        return;
+    }
+
+    var conn = this.getOutConnection();
+    if (!conn || conn.direction !== 'out') {
+        conn = this.connectTo();
+    }
+
+    this.waitForIdentified(conn, onIdentified);
+
+    function onIdentified(err) {
+        if (!err) {
+            self.nextConnAttemptDelay = 0;
+            self.nextConnAttemptTime = 0;
+            return;
+        }
+
+        if (self.nextConnAttemptDelay === 0) {
+            self.nextConnAttemptDelay = 5000;
+        } else {
+            self.nextConnAttemptDelay *= 2;
+        }
+
+        var now = Date.now();
+        if (self.nextConnAttemptTime < now) {
+            // When in the past set next attempt to now + delay
+            self.nextConnAttemptTime = now + self.nextConnAttemptDelay;
+        } else {
+            // When in the future; go further in the future
+            self.nextConnAttemptTime += self.nextConnAttemptDelay;
+        }
+    }
 };
 
 TChannelPeer.prototype.waitForIdentified =
--- a/peer.js
+++ b/peer.js
@@ -39,6 +39,8 @@ var PeerDrain = require('./drain.js').PeerDrain;
 var ObjectPool = require('./lib/object_pool');
 
 var DEFAULT_REPORT_INTERVAL = 1000;
+var INITIAL_CONN_ATTEMPT_DELAY = 5000;
+var CONN_ATTEMPT_DELAY_MULTIPLER = 2;
 
 /*eslint max-statements: [2, 40]*/
 function TChannelPeer(channel, hostPort, options) {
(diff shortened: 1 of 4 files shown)


test/peer-to-peer-load-balance.js
```diff
@@ -274,6 +275,10 @@ allocCluster.test('p2p requests where half of servers down', {
         });
         cassert.report(assert, 'expected request distribution to be ok');
 
+        var logs = cluster.logger.items();
+        assert.ok(logs.length <= 280 + 160,
```
Raynos: This number used to be ~1500-1800 before this change.

Now its ~200-350

peer.js
```diff
@@ -68,6 +68,9 @@ function TChannelPeer(channel, hostPort, options) {
     this.boundOnPendingChange = onPendingChange;
     this.scoreRange = null;
 
+    this.nextConnAttemptTime = 0;
+    this.nextConnAttemptDelay = 0;
```
kriskowal: Should these values be reset when a connection is successfully identified? Peers are long lived. If these values only grow, won’t connecting to that peer become permanently sluggish?

jcorbin:
Man... can't we just steal the connection smearing logic from hyperbahn?
- liberate IntervalScan (copy-n-paste... or spin off into separate module)
- carry a single connector on the root channel
- if choosing an interval is tricky, we could even pivot IntervalScan during its liberation to support immediate batch processing (setImmediate to kick off a run if not already running, other wise make a note that another immediate run is needed after the current one finishes)

I mean, this code you've written here probably works and all, so don't let me stop you from shipping incremental improvement, but I really think there's a better way than what we've been going towards with this feature...

peer.js
arkajit: Is `nextConnAttemptDelay` in ms? Can you add a comment clarifying?

peer.js
```diff
@@ -410,6 +413,45 @@ TChannelPeer.prototype.connectTo = function connectTo() {
...
+    this.waitForIdentified(conn, onIdentified);
+
+    function onIdentified(err) {
+        if (!err) {
+            self.nextConnAttemptDelay = 0;
+            self.nextConnAttemptTime = 0;
+            return;
+        }
+
+        if (self.nextConnAttemptDelay === 0) {
```
arkajit: Question: since the state about connection attempt delays is stored at the channel level, does this mean that failure to connect to one peer A will also cause backoff from trying to connect to a different peer B?

Wouldn't we want to backoff only in trying to connect to the downed peer and not all peers? Maybe would we could store the delay state at the subchannel level to achieve this?

peer.js
Raynos: I believe I reset these values on success.

Raynos:
@jcorbin 

I think this feature is very different from intervalScan. What I want here is backoff. I want to avoid continuously trying to connect to a dead peer.

The general implementation of this is health management and probing in the unhealthy state and backoff on the probing.

kriskowal:
:ok: Found the reset logic.

peer.js
Raynos: The state is stored on a `peer` instance. There is one peer per remote `hostPort`.

peer.js
```diff
@@ -410,6 +413,45 @@ TChannelPeer.prototype.connectTo = function connectTo() {
...
+            self.nextConnAttemptDelay = 5000;
+        } else {
+            self.nextConnAttemptDelay *= 2;
```
jcorbin: should add fuzz; should make the multiplier a constant

peer.js
```diff
@@ -410,6 +413,45 @@ TChannelPeer.prototype.connectTo = function connectTo() {
...
+
+    function onIdentified(err) {
+        if (!err) {
+            self.nextConnAttemptDelay = 0;
+            self.nextConnAttemptTime = 0;
+            return;
+        }
+
+        if (self.nextConnAttemptDelay === 0) {
+            self.nextConnAttemptDelay = 5000;
```
jcorbin: limit should be a constant

peer.js
```diff
@@ -410,6 +413,45 @@ TChannelPeer.prototype.connectTo = function connectTo() {
     return self.connect(true);
 };
 
+TChannelPeer.prototype.tryConnect = function tryConnect() {
+    var self = this;
+
+    var rightNow = Date.now();
```
jcorbin: just rename now and rightNow to be more descriptive like "tryTime", "startTime" or whatever you want, but "rightNow" and "now" are no longer serving.

peer.js
```diff
@@ -410,6 +417,47 @@ TChannelPeer.prototype.connectTo = function connectTo() {
...
+            self.nextConnAttemptTime = 0;
+            return;
+        }
+
+        if (self.nextConnAttemptDelay === 0) {
+            self.nextConnAttemptDelay = INITIAL_CONN_ATTEMPT_DELAY;
+        } else {
+            self.nextConnAttemptDelay *= CONN_ATTEMPT_DELAY_MULTIPLER;
+            // Add some amount of fuzz, +-100ms
+            self.nextConnAttemptDelay += Math.floor(100 * (Math.random() - 0.5));
```
jcorbin: fuzz percent should be a constant... also 50% is quite a lot for fuzzing, isn't 10% more typical?

jcorbin: oh ic, you're doing +/- 50ms... usually you'd just `x *= (1 + P * (Math.random() - 0.5))` for say P = 0.1 to get +/- 5% (aka 10% fuzz)...

Raynos: I copied and pasted fuzz from another part of the code base XD

PR #311, opened 2016-03-21
lazy_relay: read res code, not res flags

Raynos:
There is a bug where we read the resp flags, not the resp code.

This means our `app-error` stat is really a fragmentation stat
and not a thrift exception stat.

r: @kriskowal @jcorbin @blampe


kriskowal:
:ok:

PR #312, opened 2016-03-23
upgrade thriftrw@3.4.1

Raynos:
This fixes a bug with lists of lists in binary encode/decode
in thriftrw

r: @jcorbin @kriskowal


blampe:
:ship:

PR #313, opened 2016-03-23
Upgrade thriftrw and bufrw

kriskowal:
This brings our versions of bufrw and thriftrw into alignment, fixing the underlying problem with SeriesRW, which broke lists of lists.

r @jcorbin @Raynos


Raynos:
lgtm.

PR #320, opened 2016-04-07
Allow dynamically setting trace sample

thecadams:
This will allow us to dynamically adjust this from inside the jaeger tracing library, when we fetch a new sample rate.

### DIFF
--- a/test/trace/basic_server.js
+++ b/test/trace/basic_server.js
@@ -163,3 +163,17 @@ test('basic tracing test', function (assert) {
         }, 10);
     });
 });
+
+test('modifying traceSample', function (assert){
+    var channel = new TChannel();
+    var subservice = channel.makeSubChannel({
+        serviceName: 'subservice'
+    });
+    assert.equal(channel.traceSample, 0.01);
+    assert.equal(subservice.traceSample, 0.01);
+
+    subservice.setTraceSample(0.99);
+    assert.equal(channel.traceSample, 0.99);
+    assert.equal(subservice.traceSample, 0.99);
+    assert.end();
+});
--- a/test/trace/basic_server.js
+++ b/test/trace/basic_server.js
@@ -169,11 +169,22 @@ test('modifying traceSample', function (assert){
     var subservice = channel.makeSubChannel({
         serviceName: 'subservice'
     });
+    var subsubservice1 = subservice.makeSubChannel({
+        serviceName: 'subsubservice1'
+    });
+    var subsubservice2 = subservice.makeSubChannel({
+        serviceName: 'subsubservice2'
+    });
     assert.equal(channel.traceSample, 0.01);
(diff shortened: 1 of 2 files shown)


Raynos:
lgtm.

test/trace/basic_server.js
```diff
@@ -163,3 +163,17 @@ test('basic tracing test', function (assert) {
...
+    var channel = new TChannel();
+    var subservice = channel.makeSubChannel({
+        serviceName: 'subservice'
+    });
+    assert.equal(channel.traceSample, 0.01);
+    assert.equal(subservice.traceSample, 0.01);
+
+    subservice.setTraceSample(0.99);
+    assert.equal(channel.traceSample, 0.99);
+    assert.equal(subservice.traceSample, 0.99);
```
kriskowal: The proposed semantics appear to be that all channels share a traceSample property with their top channel. What synchronizes the traceSample across peer subchannels? Should the trace sample property only exist on the top channel and be null for other channels? Should the trace sample property of subchannels be a getter that reaches into the top channel?

kriskowal:
I have poked the tests and have further questions for @thecadams about the intended semantics.

test/trace/basic_server.js
thecadams: I can have it descend to subchannels too, that seems reasonable. In practice we want an entire app instance to share the same sample rate across both tchannel and http traces.

thecadams: The idea here is that you want a consistent sampling rate across all calls so an entire app instance will use the same sampling rate for all TChannel calls. A call to this interface should apply the samplingRate to the entire channel tree, so it does need to descend into subtrees. It also needs to guard against cycles.

kriskowal: Alternately, we can consult this.topChannel.traceSample when deciding whether to trace each request, instead of this.traceSample, and only set traceSample on the topChannel. We should fail an assertion if you call setTraceSample on any channel except the topChannel so that we avoid action at a distance confusion.

thecadams:
So I implemented recursive descent/ascent and hit an assert I wasn't aware of. It looks like TChannel only supports 2 levels of channels, which means the original implementation should be enough. I'm going to revert the recursive implementation.

https://travis-ci.org/uber/tchannel-node/jobs/121554907

kriskowal:
I am pushing for a simpler solution:

Change this line such that it always consults the topChannel.traceSample: https://github.com/uber/tchannel-node/blob/master/channel.js#L830

Change this block so it only applies traceSample to the topChannel and warn and ignore (deprecation warning) if a traceSample is provided for a subChannel. https://github.com/uber/tchannel-node/blob/master/channel.js#L218

Implement setTraceSample so that it sets traceSample, and asserts that it is only called on the topChannel.

PR #321, opened 2016-04-07
Remove ansi-color from devDependencies

Qix-:
`ansi-color` doesn't need to be there - nothing used in the module itself uses it. Noticed it when looking through dependencies and saw that `chalk` was being used instead.


Raynos:
lgtm.

PR #322, opened 2016-05-13
Work around problem with sendErrorFrame on reset lazy relay

kriskowal:
Observed an uncaught exception following this guard. Maybe we should fix the object pool.

r @raynos @jcorbin


kriskowal:
Also, @rf

Raynos:
+1

PR #323, opened 2016-05-16
Thread allowOptionalArguments through Thrift

kriskowal:
I’ve added a flag to the latest version of ThriftRW that allows users to opt-in for semantics regarding unmarked (not optional, not required) fields of argument structs that is consistent with the other ThriftRWs. We have lost the battle for required-only arguments. With the new semantics, arguments are optional by default and and you can opt-in for required.

r @jcorbin @Raynos


Raynos:
lgtm.

PR #324, opened 2016-05-27
support baseAppHeaders in as/json and as/thrift

rf:
`baseAppHeaders` allow us to specify some app headers in the call to `request`.

r @kriskowal @Raynos

### DIFF
--- a/as/json.js
+++ b/as/json.js
@@ -24,6 +24,7 @@ var Buffer = require('buffer').Buffer;
 var assert = require('assert');
 var Result = require('bufrw/result');
 var cyclicStringify = require('json-stringify-safe');
+var extend = require('xtend');
 
 var errors = require('../errors.js');
 
@@ -63,7 +64,11 @@ function send(endpoint, head, body, callback) {
     var self = this;
 
     var outreq = self.channel.request(self.reqOptions);
-    self.tchannelJSON.send(outreq, endpoint, head, body, callback);
+    var headers = head;
+    if (self.reqOptions.defaultRequestHeaders) {
+        headers = extend(self.reqOptions.defaultRequestHeaders, head);
+    }
+    self.tchannelJSON.send(outreq, endpoint, headers, body, callback);
 };
 
 TChannelJSON.prototype.request = function request(reqOptions) {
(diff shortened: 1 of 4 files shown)


test/as-json.js
```diff
@@ -78,6 +78,50 @@ allocCluster.test('getting an ok response', {
...
+
+    var server = cluster.channels[0].subChannels.server;
+    tchannelJSON.register(server, 'test', {}, testResponse);
+
+    var client = cluster.channels[1].subChannels.server;
+
+    tchannelJSON.request({
+        serviceName: 'server',
+        hasNoParent: true,
+        baseAppHeaders: {key: 'val'},
```
kriskowal: Let’s worry about this name some more. `default` seems more idiomatic than `base` (though I’m sure you considered default first), and i would like to avoid introducing the `App` abbreviation. Let’s classify headers as “transport headers”, “request headers”, and “context headers”. `defaultRequestHeaders` seems like a step better.

kriskowal:
name consideration aside, looks good to me.

as/json.js
```diff
@@ -63,7 +64,7 @@ function send(endpoint, head, body, callback) {
     var self = this;
 
     var outreq = self.channel.request(self.reqOptions);
-    self.tchannelJSON.send(outreq, endpoint, head, body, callback);
+    self.tchannelJSON.send(outreq, endpoint, extend(head, self.reqOptions.baseAppHeaders), body, callback);
```
Raynos: please avoid the `extend` call unless it exists.

Also your extending in the wrong direction...

as/thrift.js
```diff
@@ -527,7 +528,7 @@ function send(endpoint, head, body, callback) {
     var self = this;
 
     var outreq = self.channel.request(self.reqOptions);
-    self.tchannelThrift.send(outreq, endpoint, head, body, callback);
+    self.tchannelThrift.send(outreq, endpoint, extend(head, self.reqOptions.baseAppHeaders), body, callback);
```
Raynos: please avoid the `extend` call unless it exists.

Also your extending in the wrong direction...

Raynos:
lgtm.

PR #325, opened 2016-06-02
port over tchannel file watcher

Raynos:
This is a port of the FileWatcher that @dansimau added
to rt-supply.

I've cleaned it up a bit by improving the drain, adding
both sync & async reload support

Next up is :

 - Integrate into `makeSubChannel()`
 - Add tests for the file watching semantics.
 - Add tests for additions
 - Add tests for removals

r: @dansimau @syyang @rf


dansimau:
epic, thanks @Raynos

channel-watcher.js
```diff
@@ -0,0 +1,197 @@
...
+    if (!fs.existsSync(this.filePath)) {
+        assert(false, 'Peer hosts file does not exist: ' + this.filePath);
+    }
+};
+
+ChannelWatcher.prototype._readPeerListFromFileSync =
+function _readPeerListFromFileSync() {
+    var tuple = safeJSONParse(fs.readFileSync(this.filePath, 'utf8'));
+    if (tuple[0]) {
+        assert(false, 'Invalid JSON for TChannel peer host file at ' + this.filePath);
```
BinChang: this could be dangerous,  in history,  we had outage because similar logic.   If the host-file update is not atomic,  i.e., not the file is updated chunk-by-chunk, and you will get events for a file that contains invalid JSON content,  this line might kill the whole node worker.  


@jwyuan and Sebastian Yates know the context.

channel.js
```diff
@@ -603,6 +604,11 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
             }
         }
     }
+    if (options.hostsWatchFilePath) {
```
dansimau: Should `hostsWatchFilePath` actually be `peerFile`?

The option for providing a list of peers without a file is `peers`. I think the main theme of the top-level option is that we're providing a mechanism to load a list of peers from a file. The fact that we watch the file for changes is just a nice detail. It doesn't need to be exposed in the name.

channel.js
```diff
@@ -67,6 +67,7 @@ var RetryFlags = require('./retry-flags.js');
 var TimeHeap = require('./time_heap');
 var CountedReadySignal = require('ready-signal/counted');
 var BatchStatsd = require('./lib/statsd.js');
+var ChannelWatcher = require('./channel-watcher.js');
```
dansimau: I think a more apt name would be `PeerFileWatcher`.

channel-watcher.js
```diff
@@ -0,0 +1,197 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var safeJSONParse = require('safe-json-parse/tuple');
+var fs = require('fs');
+var assert = require('assert');
+
+function ChannelWatcher(channel, opts) {
```
dansimau: I think a more apt name would be `PeerFileWatcher`.

channel-watcher.js
mbrysa: Yes, we're guarding against it in the processes that generate the hostfiles by atomically moving a finished file over the old one.
But this should not be a fatal error, I'll make it a warning.

channel.js
mbrysa: Yes, that's a much better name.

channel-watcher.js
mbrysa: Yes, I'll make it so.

channel-watcher.js
Raynos: This is only used synchronously on startup.
This should be handled or stay fatal. I don't know how to handle this error.
Your process will start up with an empty peers list and all outbound requests will fail. 

That being said we can add a `hardFail` option that defaults to false. If you set `hardFail` to `false` then we assume that this sub channel is non critical and that the service can run in a partially available or partially degraded state.

If this sub channel is critical ( e.g. cannot make trips without talking to this service ) then you should opt into `hardFail` and let the process exit / restart / fail to join the ringpop ring / etc.

channel-watcher.js
Raynos: done.

channel.js
Raynos: done.

peer-file-watcher.js
```diff
@@ -0,0 +1,197 @@
...
+
+PeerFileWatcher.prototype.updatePeers =
+function updatePeers(newPeers) {
+    // Take a snapshot of current, existing peers. This is used to delete old
+    // peers later.
+    var oldPeers = this.channel.peers.keys().slice();
+
+    var i;
+
+    // Load new peers; duplicates are ignored
```
syyang: just to double check, is the comment accurate? where are the duplicates ignored?

Raynos: `peers.add(hostPort);` ( https://github.com/uber/tchannel-node/blob/master/sub_peers.js#L68-L75 ) is idempotent.

syyang: got it.

syyang:
lgtm. thanks!

Raynos:
Published 3.7.3

PR #326, opened 2016-06-03
Add free guard to LazyRelayInReq#onTimeout

rf:
r @raynos @kriskowal

Just adding a guard so this doesn't keep causing uncaughts and alerting


kriskowal:
LG™

PR #328, opened 2016-07-01
timeout should hold until last byte

Raynos:
We've had an issue with the `rt-msg` service in production
where they were seeing multi-minute latencies, probably
because the timeout measure the time until the first call
response frame, not the last.

We had previously supported streaming by making the timeout
behave until the first call response frame.

However, this causes issues for people that encounter
fragmented call responses.

Since we have barely any streaming customers we should
recommend that people who use streaming and want timeout
until first response frame simply set a very large timeout
and a secondary custom timeout until the first response.

r: @ssyang @jcorbin @kriskowal

### DIFF
--- a/in_request.js
+++ b/in_request.js
@@ -169,7 +169,10 @@ TChannelInRequest.prototype.onTimeout = function onTimeout(now) {
     var self = this;
     var timeoutError;
 
-    if (!self.res || self.res.state === States.Initial) {
+    if (!self.res ||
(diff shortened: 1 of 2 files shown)


jcorbin:
Sounds sensible, we can modify with an optional header as needed when needed.

syyang:
lgtm

Raynos:
Published 3.8.0

PR #329, opened 2016-07-26
add a isUnhealthyError method into TChannel used for client-side throttling

zhijinli:
@Raynos suggested to do this so we don't have to import the tchannel dependency into our libs repo @syyang  @charliezhang


Raynos:
lgtm, please restart travis until tests pass :)

PR #330, opened 2016-08-05
Improve the connection backoff logic.

Raynos:
It looks like in production we backoff far too aggressively.

This is caused by having multiple concurrent `tryConnect()` attempts
which will all "succeed" or "fail" and thus apply the doubling
logic many times ballooning the retry attempt delay.

I've also added a hard gap of 30s backoff so every 30s we will
try to open the connection.

r: @syyang


syyang:
lgtm. thanks.

blampe:
Does this need a test?

Raynos:
We are going to test & verify this in production, will merge once verified.

Raynos:
Verified in prod. Will land once green.

PR #334, opened 2016-08-20
introducing openTracer without removing old-style tracing

badiib:
adding a subTracer in the tracer object in channel called openTracer. This means each request needs to have a secondary span object that's called openSpan.

This diff is to define the cut-points for openTracer without removing any parts of the old style tracing.

Still pending:
1. Proper creation of spans given the head k,v map.
2. New tests for thrift and json tracing.

### DIFF
--- a/as/json.js
+++ b/as/json.js
@@ -91,7 +91,7 @@ TChannelJSON.prototype.send = function send(
 ) {
 
     var self = this;
-    req.openSpan = req.channel.startSpan(head);
+    req.channel.startSpan(req, head);
 
     if (!self.logger) {
         self.logger = req.channel.logger;
@@ -194,7 +194,7 @@ TChannelJSON.prototype.register = function register(
         }
 
         var v = parseResult.value;
-        req.openSpan = req.channel.startSpan(v.head);
+        req.channel.startSpan(req, v.head);
         handlerFunc(opts, req, v.head, v.body, onResponse);
 
         function onResponse(err, respObject) {
--- a/as/json.js
+++ b/as/json.js
@@ -122,7 +122,7 @@ TChannelJSON.prototype.send = function send(
     );
 
     function onResponse(err, resp, arg2, arg3) {
-        req.channel.finishSpan(req.openSpan, err);
+        req.channel.finishSpan(req, err);
         if (err) {
             return callback(err);
         }
@@ -198,7 +198,7 @@ TChannelJSON.prototype.register = function register(
         handlerFunc(opts, req, v.head, v.body, onResponse);
 
         function onResponse(err, respObject) {
-            req.channel.finishSpan(req.openSpan, err);
+            req.channel.finishSpan(req, err);
             if (err) {
                 self.logger.error('Got unexpected error in handler', {
                     endpoint: arg1,
--- a/as/json.js
+++ b/as/json.js
@@ -91,6 +91,7 @@ TChannelJSON.prototype.send = function send(
 ) {
 
     var self = this;
+    req.openSpan = req.channel.startSpan(head);
 
     if (!self.logger) {
         self.logger = req.channel.logger;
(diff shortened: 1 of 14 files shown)


channel.js
```diff
@@ -1081,4 +1082,22 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
     return errors.isUnhealthy(codeName);
 };
 
+// TODO - make this do smart things
+TChannel.prototype.startSpan = function startSpan(head) {
```
yurishkuro: you probably need to pass `req`, not just `head`, because you want to start the new span as a child of the existing span, which will be stored in the `req`

as/json.js
```diff
@@ -91,6 +91,7 @@ TChannelJSON.prototype.send = function send(
 ) {
 
     var self = this;
+    req.openSpan = req.channel.startSpan(head);
```
yurishkuro: nit: if you pass `req` instead, the method can store the span on `req` instead of returning

channel.js
```diff
@@ -1081,4 +1082,22 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
...
+    var self = this;
+    if (self.tracer) {
+        return self.tracer.openTracer.startSpan('TODO');
+    }
+    return null;
+};
+
+TChannel.prototype.finishSpan = function startSpan(openSpan, err) {
```
yurishkuro: typo: function name is `finishSpan`

in_response.js
```diff
@@ -49,6 +49,7 @@ function TChannelInResponse(id, options) {
     self.headers = options.headers || {};
     self.ok = self.code === 0; // TODO: probably okay, but a bit jank
     self.span = options.span || null;
+    self.openSpan = options.openSpan || null;
```
yurishkuro: `|| null` seems redundant

yurishkuro:
Main suggestion is to minimize the exposure outside of the new methods on the channel, i.e. pass request to them, that way the code outside doesn't even need to know about openSpan field.

as/json.js
```diff
@@ -192,9 +194,11 @@ TChannelJSON.prototype.register = function register(
         }
 
         var v = parseResult.value;
+        req.channel.startSpan(req, v.head);
         handlerFunc(opts, req, v.head, v.body, onResponse);
 
         function onResponse(err, respObject) {
+            req.channel.finishSpan(req.openSpan, err);
```
yurishkuro: nit: if it was me, I would keep this symmetric with startSpan and pass `(req, err)`, without revealing there's such a thing as `openSpan`

channel.js
```diff
@@ -1081,4 +1082,21 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
...
 
+// TODO - make this do smart things
+TChannel.prototype.startSpan = function startSpan(req, head) {
+    var self = this;
+    if (self.tracer) {
+        req.openSpan = self.tracer.openTracer.startSpan('TODO');
+    }
+};
+
+TChannel.prototype.finishSpan = function endSpan(openSpan, err) {
```
yurishkuro: function name mismatch: finishSpan / endSpan

trace/agent.js
```diff
@@ -42,6 +43,8 @@ function Agent(options) {
     if (options.reporter) {
         this.reporter = options.reporter;
     }
+
+    this.openTracer = options.openTracer || openTracing;
```
yurishkuro: is this needed?  I thought openTracer is defined on the channel. This agent could stay the old way

badiib: this is only for testing purposes to ensure openTracer is non-null

Raynos:
The `opentracing` library still needs to meet our module requirements. It should be authored in ES5 instead of ES6.

We will also need to benchmark performance.

I recommend you land this onto a "tracing" branch instead of "master" branch.

We used to have a tracing benchmark ( https://github.com/uber/tchannel-node/blob/master/benchmarks/Makefile#L46 ) that may be broken on master / need some parameters tweaked, but it should allow us to verify performance. The benchmark has tracing support ( https://github.com/uber/tchannel-node/blob/master/benchmarks/multi_bench.js#L161-L176 ) and will need tweaking to support writing to jeager / opentracing.

PR #336, opened 2016-09-07
Fixup benchmarks Makefile to work out of box.

Raynos:
The Makefile is not configured to work out of the box.

There was also one bug in the benchmark bootstrapper that
caused the tracing benchmark to fail

I've also turned on 100% trace reporting instead of 1% for the
tracing benchmark...

So currently when benchmarking wrt to tracing:

 - `make take` will benchmark p2p, with span creation and noop-reporter
 - `make take_relay` will benchmark with relay, with span creation & noop-reporter
 - `make trace` will benchmark with relay, with span creation & 100% reporter

The four modes do the following on my laptop,

 - 6.5k @ `make take`, hand-edited to have trace: false, aka no span creation
 - 5k @ `make take`, normal, span creation, noop reporter
 - 3k @ `make take_relay`, normal, span creation, relay, noop reporter
 - 1k @ `make take_trace`, normal, span creation, relay, 100% tcollector reporter

We currently do not have a `make take_no_trace` so you have to hand edit
and s/trace: true/trace: false/.

r: @rf @oibe

### DIFF
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -8,7 +8,7 @@ torch-client-only-bench:
 
 torch-server-bench:
 	node index.js --relay --torch server --torchFile ./flame.raw --torchTime 10 \
-		-- --relay --skipPing -s 4096,16384 -p 1000,10000
+		-- --relay --skipPing -s 4096,16384 -p 1000
 
 torch-relay-bench:
 	node index.js --relay --torch relay --torchFile ./flame.raw --torchTime 10 \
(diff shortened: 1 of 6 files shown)


oibe:
@Raynos Thanks a lot for the analysis.  lgtm.  I will use this branch as a starting point for finishing up the opentracing diff that Rooz started.

Raynos:
All files effected are in the `benchmarks/` folder so I'm going to go ahead and land this.

PR #337, opened 2016-09-08
support per-client retry budget to avoid excessive retries overwhelming server

zhijinli:
@Raynos @syyang @hanjunx @charliezhang PTAL

### DIFF
--- a/channel.js
+++ b/channel.js
@@ -56,6 +56,7 @@ var TChannelServiceNameHandler = require('./service-name-handler');
 var errors = require('./errors');
 var EventEmitter = require('./lib/event_emitter.js');
 var ObjectPool = require('./lib/object_pool');
+var RetryRatioTracker = require('./lib/retry_ratio_tracker');
 
 var TChannelAsThrift = require('./as/thrift');
 var TChannelAsJSON = require('./as/json');
@@ -86,6 +87,9 @@ var MAX_TOMBSTONE_TTL = 5000;
 // TODO restore spying
 // var Spy = require('./v2/spy');
 // var dumpEnabled = /\btchannel_dump\b/.test(process.env.NODE_DEBUG || '');
+
+var DEFAULT_ENABLE_MAX_RETRY_RATIO = false;
+var DEFAULT_MAX_RETRY_RATIO = 0.1;
 
 /*eslint-disable max-statements*/
 function TChannel(options) {
@@ -281,6 +285,13 @@ function TChannel(options) {
             return;
         }
         self.sanityTimer = self.timers.setTimeout(doSanitySweep, SANITY_PERIOD);
+    }
+
+    this.enableMaxRetryRatio = this.options.enableMaxRetryRatio || DEFAULT_ENABLE_MAX_RETRY_RATIO;
+    if (this.enableMaxRetryRatio) {
+        this.maxRetryRatio = this.options.maxRetryRatio || DEFAULT_MAX_RETRY_RATIO;
+        this.retryRatioTracker = this.topChannel ?
+            RetryRatioTracker({ timers: this.timers }) : null;
     }
 }
 inherits(TChannel, EventEmitter);
@@ -945,6 +956,10 @@ TChannel.prototype.close = function close(callback) {
         self.sanityTimer = null;
     }
 
+    if (self.retryRatioTracker) {
+        self.retryRatioTracker.destroy();
+    }
+
     if (self.serverSocket) {
         ++counter;
         if (self.serverSocket.address()) {
@@ -1083,5 +1098,6 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
     var codeName = errors.classify(err);
     return errors.isUnhealthy(codeName);
 };
+
 
 module.exports = TChannel;
--- a/channel.js
+++ b/channel.js
@@ -289,10 +289,11 @@ function TChannel(options) {
 
     // for client retry budget
     this.enableMaxRetryRatio = this.options.enableMaxRetryRatio || DEFAULT_ENABLE_MAX_RETRY_RATIO;
-    this.maxRetryRatio = this.options.maxRetryRatio || DEFAULT_MAX_RETRY_RATIO;
+    this.maxRetryRatio = this.options.maxRetryRatio || DEFAULT_MAX_RETRY_RATIO; // retries to requests
+    assert(this.maxRetryRatio >= 0.0, 'maxRetryRatio must be non-negative');
     this.retryRatioTracker = null;
     if (this.enableMaxRetryRatio && this.topChannel) { // only track ratio in sub channel
-        this.retryRatioTracker = RetryRatioTracker({
+        this.retryRatioTracker = new RetryRatioTracker({
             rateCounterInterval: this.options.rateCounterInterval,
             rateCounterNumOfBuckets: this.options.rateCounterNumOfBuckets,
             timers: this.timers
@@ -615,6 +616,10 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
 
     opts.topChannel = self;
     opts.timeHeap = self.timeHeap;
+
+    opts.enableMaxRetryRatio = options.enableMaxRetryRatio;
+    opts.maxRetryRatio = options.maxRetryRatio;
+
     var chan = TChannel(opts);
 
     if (options.peers) {
--- a/channel.js
+++ b/channel.js
@@ -287,11 +287,16 @@ function TChannel(options) {
         self.sanityTimer = self.timers.setTimeout(doSanitySweep, SANITY_PERIOD);
     }
 
+    // for client retry budget
     this.enableMaxRetryRatio = this.options.enableMaxRetryRatio || DEFAULT_ENABLE_MAX_RETRY_RATIO;
-    if (this.enableMaxRetryRatio) {
-        this.maxRetryRatio = this.options.maxRetryRatio || DEFAULT_MAX_RETRY_RATIO;
-        this.retryRatioTracker = this.topChannel ?
-            RetryRatioTracker({ timers: this.timers }) : null;
+    this.maxRetryRatio = this.options.maxRetryRatio || DEFAULT_MAX_RETRY_RATIO;
+    this.retryRatioTracker = null;
+    if (this.enableMaxRetryRatio && this.topChannel) { // only track ratio in sub channel
+        this.retryRatioTracker = RetryRatioTracker({
+            rateCounterInterval: this.options.rateCounterInterval,
+            rateCounterNumOfBuckets: this.options.rateCounterNumOfBuckets,
+            timers: this.timers
+        });
     }
 }
 inherits(TChannel, EventEmitter);
@@ -1098,6 +1103,5 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
     var codeName = errors.classify(err);
     return errors.isUnhealthy(codeName);
 };
-
 
 module.exports = TChannel;
--- a/lib/rate_counter.js
+++ b/lib/rate_counter.js
@@ -30,20 +30,17 @@ var DEFAULT_NUM_OF_BUCKETS = 20;
 function RateCounter(options) {
     assert(options.timers, 'options.timers required');
 
-    if (!(this instanceof RateCounter)) {
-        return new RateCounter(options);
-    }
-
     var self = this;
-    self.index = 0;
-    self.rate = 0; // requests per interval
-    self.numOfBuckets = options.numOfBuckets || DEFAULT_NUM_OF_BUCKETS;
-    self.buckets = [];
-    self.buckets[0] = 0;
 
-    self.timers = options.timers;
-    self.rateInterval = options.rateInterval || DEFAULT_RATE_INTERVAL;
-    self.refreshInterval = self.rateInterval / self.numOfBuckets;
+    this.index = 0;
+    this.rate = 0; // requests per interval
+    this.numOfBuckets = options.numOfBuckets || DEFAULT_NUM_OF_BUCKETS;
+    this.buckets = [];
+    this.buckets[0] = 0;
+
+    this.timers = options.timers;
+    this.rateInterval = options.rateInterval || DEFAULT_RATE_INTERVAL;
+    this.refreshInterval = this.rateInterval / this.numOfBuckets;
 
     _refreshAndScheduleNext();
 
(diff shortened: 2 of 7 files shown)


channel.js
```diff
@@ -1084,4 +1096,30 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
...
+        incAndScheduleDec('retryRateCounter');
+    }
+    incAndScheduleDec('requestRateCounter');
+
+    function incAndScheduleDec(counterName) {
+        if (typeof self[counterName] !== 'number') {
+            return;
+        }
+        self[counterName]++;
+        self.timers.setTimeout(function offset() {
```
syyang: is scheduling timer for every request performant when you have high QPS?

request.js
```diff
@@ -459,6 +464,10 @@ TChannelRequest.prototype.shouldRetryError = function shouldRetryError(err) {
         return false;
     }
 
+    if (self.channel.currentRetryRatio() > self.maxRetryRatio) {
+        return false;
```
syyang: can we log some stats that are rejected due to retry budget?

channel.js
zhijinli: not sure, @Raynos what do you think? can/should we benchmark this?

channel.js
```diff
@@ -282,6 +284,11 @@ function TChannel(options) {
         }
         self.sanityTimer = self.timers.setTimeout(doSanitySweep, SANITY_PERIOD);
     }
+
+    // for per-client retry budget
+    this.retryRateCounter = 0;
```
Raynos: We probably want a different retry budget for each sub channel.

channel.js
```diff
@@ -810,6 +820,7 @@ function fastRequestDefaults(reqOpts) {
 function RequestDefaults(reqDefaults) {
     this.timeout = reqDefaults.timeout || 0;
     this.retryLimit = reqDefaults.retryLimit || 0;
+    this.maxRetryRatio = reqDefaults.maxRetryRatio || 1.0;
```
Raynos: Do we really want this parameter per request ? I suspect that this would be something configured for the entire serviceName.

channel.js
```diff
@@ -1084,4 +1096,30 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
...
+
+TChannel.prototype.incrementRequest = function incrementRequest(isRetry) {
+    var self = this;
+
+    if (isRetry) {
+        incAndScheduleDec('retryRateCounter');
+    }
+    incAndScheduleDec('requestRateCounter');
+
+    function incAndScheduleDec(counterName) {
```
Raynos: avoid closures, this should just be another method.

channel.js
```diff
@@ -1084,4 +1096,30 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
...
+        if (typeof self[counterName] !== 'number') {
+            return;
+        }
+        self[counterName]++;
```
Raynos: avoid dynamic field lookup, just have a static switch.

channel.js
Raynos: This will have severe performance overhead.

channel.js
```diff
@@ -1084,4 +1096,30 @@ TChannel.prototype.isUnhealthyError = function isUnhealthyError(err) {
...
+        self.timers.setTimeout(function offset() {
+            self[counterName]--;
```
Raynos: We can just use a different data structure like a rolling bucket, for example: https://github.com/uber/hyperbahn/blob/master/rate_limiter.js#L36-L74

request.js
```diff
@@ -79,6 +80,7 @@ inherits(TChannelRequest, EventEmitter);
 
 TChannelRequest.defaultRetryLimit = 5;
 TChannelRequest.defaultTimeout = 100;
+TChannelRequest.defaultMaxRetryRatio = 1.0;
```
Raynos: the default should be `null` ? That's backwards compatibility, alternatively the default can be `5.0` to match the `defaultRetryLimit`

request.js
```diff
@@ -273,6 +275,9 @@ function emitOutboundCallsSent() {
 TChannelRequest.prototype.resend = function resend() {
     var self = this;
 
+    var isRetry = self.resendSanity < self.limit;
+    self.channel.incrementRequest(isRetry);
```
Raynos: This `channel` is the sub channel so this does mean we have on set of retry limits per outbound thing.

However just for sanity, let's move the logic into a new class and do `self.channel.x.increment()`

Then we can only allocate it conditionally like ( https://github.com/uber/tchannel-node/blob/acc7e9b126e6eba4d30ec9edaa8cd74a2f70cf70/channel.js#L187-L191 ). We can have it be `null` on the top channel.

request.js
Raynos: by default every tchannel will emit metrics for outbound requests, it will emit the amount of attempts and success as well as the amount of retries. 

In theory we can return a custom error class here and then we should have an operational-error stat with the type of error in the name.

Raynos: Aha, currently tchannel will emit the "last failure" when it cannot retry again.

So the way to have visibility without adding new stats is to look at the retries graph and see if it flat lines at the 10% cap.

If you look at success, error & retries then in theory you should see something like :

 - 100% success,
 - 100% success, 2% retries,
 - 99% success, 5% retries, 1% error
 - 98% success, 10% retries, 2% error,
 - 88% success, 10% retries, 12% errors

However we can emit new metrics for retry bailout. That's probably a good idea.

channel.js
zhijinli: already done

channel.js
zhijinli: removed, and made it a config for channel only

channel.js
zhijinli: done

channel.js
zhijinli: gone after refactored

channel.js
zhijinli: neat! thanks for the suggestion

request.js
zhijinli: removed along with removing the param from request

request.js
zhijinli: done

request.js
zhijinli: I think emitting a new metric is the most straightforward way to see how this feature is performing

channel.js
zhijinli: nice catch, thanks for pointing out.

channel.js
```diff
@@ -282,6 +286,13 @@ function TChannel(options) {
         }
         self.sanityTimer = self.timers.setTimeout(doSanitySweep, SANITY_PERIOD);
     }
+
+    this.enableMaxRetryRatio = this.options.enableMaxRetryRatio || DEFAULT_ENABLE_MAX_RETRY_RATIO;
+    if (this.enableMaxRetryRatio) {
+        this.maxRetryRatio = this.options.maxRetryRatio || DEFAULT_MAX_RETRY_RATIO;
```
Raynos: we should not conditionally assign fields in the constructor as that de-optimizes the code.

Please pre-assign them outside the if statement

```
this.maxRetryRation = DEFAULT_MAX_RETRY_RATIO;
this.retryRatioTracker = null;
if (this.enableMaxRetryRatio) {
  ...
}
```

request.js
```diff
@@ -273,6 +273,11 @@ function emitOutboundCallsSent() {
 TChannelRequest.prototype.resend = function resend() {
     var self = this;
 
+    if (self.channel.enableMaxRetryRatio) {
+        var isRetry = self.resendSanity < self.limit;
```
Raynos: This condition is quite difficult to understand;

Try:

```
var isRetry = self.outReqs.length > 0;
```

request.js
```diff
@@ -459,6 +464,22 @@ TChannelRequest.prototype.shouldRetryError = function shouldRetryError(err) {
         return false;
     }
 
+    if (self.channel.enableMaxRetryRatio &&
+        self.channel.retryRatioTracker.currentRetryRatio() > self.channel.maxRetryRatio) {
```
Raynos: whitespace:

````
if (a &&
    b &&
    c // new line here
) {
    /* block */
}

File: lib/retry_ratio_tracker.js [hunk 13]
```diff
@@ -0,0 +1,77 @@
...
+    if (isRetry) {
+        this.retryRateCounter.increment();
+    }
+};
+
+RetryRatioTracker.prototype.refreshRateCounters = function refreshRateCounters() {
+    var self = this;
+    self.requestRateCounter.refresh();
+    self.retryRateCounter.refresh();
+    this.refreshCountersTimer = this.timers.setTimeout(function refresh() {
```
Raynos: pre-allocate this closure in the constructor once instead of allocating it every 50ms.

channel.js
zhijinli: done

lib/retry_ratio_tracker.js
```diff
@@ -0,0 +1,77 @@
...
+
+RetryRatioTracker.prototype.currentRetryRatio = function currentRetryRatio() {
+    if (this.requestRateCounter.rps === 0) {
+        return 0;
+    }
+    return 1.0 * this.retryRateCounter.rps / this.requestRateCounter.rps;
+};
+
+RetryRatioTracker.prototype.incrementRequest = function incrementRequest(isRetry) {
+    this.requestRateCounter.increment();
```
Raynos: I dont think we want to increment this if its a retry.

e.g. if we do 500 RPS and we want 0.1 retry budget then thats 50 retries, but if you increment for retries then we will be down 550 RPS * 0.1; aka 55 retries.

This will get worse for example retry budget of 0.5 which would turn into retry budget of 0.75

request.js
zhijinli: done

lib/retry_ratio_tracker.js
```diff
@@ -0,0 +1,77 @@
...
+    this.timers = options.timers;
+    this.retryRateCounter = RateCounter(rateCounterOpts);
+    this.requestRateCounter = RateCounter(rateCounterOpts);
+    this.refreshCounterInterval = RATE_INTERVAL / numOfBuckets;
+
+    this.refreshRateCounters();
+}
+
+RetryRatioTracker.prototype.currentRetryRatio = function currentRetryRatio() {
+    if (this.requestRateCounter.rps === 0) {
```
Raynos: This logic will work create once the process is stable because it will look at the last 20 * 50ms buckets for both data.

However it may behave weird around startup time.

For example a service might do 100 QPS, but in the first 50ms we may do 5 requests and try to do 4 retries, with 5 requests we are allowed 0.5 retries but we actually have budget for 10 retries.

Given the implementation of `this.requestRateCounter` we can actually add another gaurd.

```
// Whilst the request rate counter is not yet warmed up, allow ALL retries
if (this.requestRateCounter.buckets.length < numOfBuckets) {
  return 0;
}
```

I don't know what the best approach is here to be honest in the warmup phase, maybe something more dump like 

```
// Whilst the request rate counter is not warmed up, allow a "free" 10 retries
if (this.requestRateCounter.buckets.length < numOfBuckets &&
    this.retryRateCounter.rps < 10
) {
  return 0;
}
```

Raynos:
Finished second review.

Raynos:
After speaking with @zhijinli in person we realized that this implementation might behave poorly for any clients with <10 outbound QPS.

A maxRetryRatio of 0.1 for <10 outbound QPS rounds down to zero, aka no retries.

I speculate that configuring the RPS buckets to be 30 seconds of data instead of 1 second of data and allowed a max of 0.1 worth of retries per 30 second bucket would create a better allowance of bursty retries whilst still serving the original purpose of putting an upper bound on the amount of retries being send.

Most of our retries overload scenarios consist of multiple minutes of sustained overloaded requests.

lib/retry_ratio_tracker.js
zhijinli: Nice catch and suggestion. I'm adding a `isWarmedUp` method which is checked before retries. Not quite sure about the RPS checking part - I understand what you want to address, but I'm wondering is there a better way than directly checking RPS against a fixed value? any idea of handling it better?

lib/retry_ratio_tracker.js
zhijinli: I think it depends on how we define the retry ratio and whether RPS should include retries:

it was implemented for totalRetries/totalRequestsIncludingRetries, but looks like totalRetries/totalRequestsNotIncludingRetries (which is what you suggested) seems better, so done :)

lib/retry_ratio_tracker.js
zhijinli: done

request.js
zhijinli: done

zhijinli:
> A maxRetryRatio of 0.1 for <10 outbound QPS rounds down to zero, aka no retries.

as discussed again, this should not be the case

> I speculate that configuring the RPS buckets to be 30 seconds of data instead of 1 second of data and allowed a max of 0.1 worth of retries per 30 second bucket would create a better allowance of bursty retries whilst still serving the original purpose of putting an upper bound on the amount of retries being send.

makes sense. this might be quite helpful for our use case. I refactored `RateCounter` to make the rate interval configurable, and set 30s as the default.

lib/rate_counter.js
```diff
@@ -0,0 +1,84 @@
...
+
+var assert = require('assert');
+
+var DEFAULT_RATE_INTERVAL = 1000;
+var DEFAULT_NUM_OF_BUCKETS = 20;
+
+function RateCounter(options) {
+    assert(options.timers, 'options.timers required');
+
+    if (!(this instanceof RateCounter)) {
```
Raynos: nitpick: this gaurd is an anti pattern, remove it and ensure you invoke `new` at all allocation call sites.

lib/rate_counter.js
```diff
@@ -0,0 +1,84 @@
...
+        return new RateCounter(options);
+    }
+
+    var self = this;
```
Raynos: nitpick: `var self = this;` is an anti pattern, only use it inside closures.

Aka all top-level `self` usages can become `this`, any usages to `self` inside a closure need to keep referring to `self`.

lib/retry_ratio_tracker.js
```diff
@@ -0,0 +1,67 @@
...
+
+module.exports = RetryRatioTracker;
+
+var RateCounter = require('./rate_counter');
+
+var DEFAULT_RATE_COUNTER_INTERVAL = 30 * 1000;
+var DEFAULT_RATE_COUNTER_NUM_OF_BUCKETS = 30 * 20;
+
+function RetryRatioTracker(options) {
+    if (!(this instanceof RetryRatioTracker)) {
```
Raynos: nitpick: constructor anti pattern.

Raynos:
Other then style nitpicks this PR looks good.

PR #339, opened 2016-09-23
handle edge case where there is no connection

Raynos:
There is an edge case where a server connection closes in
the middle of a sanity sweep. This would cause an uncaught
as that connection cannot be sweeped.

If the connection doesn't exist we just move along to the next
step in sanity sweep.

r: @rf @ssyang


Raynos:
Actually, r: @syyang

rf:
test plox

syyang:
could you fix the CI? looks good otherwise.

PR #340, opened 2016-10-06
Move updatePeer and drainPeer up to TChannel level

albertyfwu:
Moving `updatePeer` and `drainPeer` up to `TChannel` level so we can update without using peer file watcher.

### DIFF
--- a/channel.js
+++ b/channel.js
@@ -769,7 +769,11 @@ TChannel.prototype.drainPeer = function drainPeer(peer) {
             );
         }
 
-        self.topChannel.peers.delete(peer.hostPort);
+        if (self.topChannel) {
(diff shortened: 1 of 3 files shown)


channel.js
```diff
@@ -712,6 +712,67 @@ TChannel.prototype.address = function address() {
...
+        if (err) {
+            self.logger.warn(
+                'TChannel: error closing peer, deleting anyhow',
+                peer.extendLogInfo(peer.draining.extendLogInfo({
+                    error: err
+                }))
+            );
+        }
+
+        self.topChannel.peers.delete(peer.hostPort);
```
Raynos: need an `if` check here.

Raynos:
lgtm.

PR #341, opened 2016-10-07
request: checkTimeout() after identified

Raynos:
We need to do another checkTimeout() after the request
is identified to double check that we did not consume our
request budget in the connection init handshake.

We also need to update the `elapsed` time so that when we
make our outbound request we set a proper `ttl-elapsed` timeout
on the per attempt request.

r: @zhijin @syyang

### DIFF
--- a/sub_peers.js
+++ b/sub_peers.js
@@ -35,6 +35,7 @@ function TChannelSubPeers(channel, options) {
     var self = this;
     this.peerScoreThreshold = this.options.peerScoreThreshold || 0;
     this.choosePeerWithHeap = channel.choosePeerWithHeap;
+    this.socketInitTimeout = channel.initTimeout;
 
     this.hasMinConnections = typeof options.minConnections === 'number';
     this.minConnections = options.minConnections;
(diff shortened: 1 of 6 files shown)


syyang:
could we add a unit test please?

zhijinli:
LGTM, thanks Jake!

Raynos:
Now with tests.

syyang:
lgmt. thanks!

sub_peers.js
```diff
@@ -35,6 +35,7 @@ function TChannelSubPeers(channel, options) {
     var self = this;
     this.peerScoreThreshold = this.options.peerScoreThreshold || 0;
     this.choosePeerWithHeap = channel.choosePeerWithHeap;
+    this.socketInitTimeout = channel.initTimeout;
```
syyang: per offline discussion with jake, this channel is a sub channel, and points to the initTimeout provided by client.

syyang:
thanks!

sub_peers.js
Raynos: Because `Peer` and `Connection` class are managed by `rootChannel`.

The best way to apply sub channel options is in the `SubPeers` class, utilizing mutation.

blampe:
I the future please update the changelog for changes like this.

PR #343, opened 2016-10-25
ensure that requests timeout even when initOp

Raynos:
This improves the pre-flight waitForIdentified() logic
in the tchannel.Request class.

Here we will actually do a check to see if we have
a usable connection and skip the waiting.

If we have to wait because there are zero connections
we also allocate a RequestOperation with a properly
configured timeout and add it to the timeHeap.

This means that the request will timeout within it's
specified `timeout` irregardless of what the connection
initTimeout value is.

We clean up and cancel the timeout once a connection
is identified and available

r: @zhijin @syyang

### DIFF
--- a/request.js
+++ b/request.js
@@ -84,13 +84,13 @@ function RequestOperation(req, timeout, peer, waitForIdentifiedSlot) {
     this.req = req;
     this.timeout = timeout;
     this.peer = peer;
-    this.waitForIdentifiedSlot = waitForIdentifiedSlot;
+    this.waitForIdentifiedSlot =
+        typeof waitForIdentifiedSlot === 'number' ?
+        waitForIdentifiedSlot : -1;
 }
 
 RequestOperation.prototype.onTimeout = function onTimeout(now) {
-    if (this.waitForIdentifiedSlot !== null &&
-        this.waitForIdentifiedSlot !== -1
-    ) {
+    if (this.waitForIdentifiedSlot !== -1) {
         this.peer.stopWaitingForIdentified(this.waitForIdentifiedSlot);
     }
 
--- a/request.js
+++ b/request.js
@@ -80,6 +80,23 @@ inherits(TChannelRequest, EventEmitter);
 TChannelRequest.defaultRetryLimit = 5;
 TChannelRequest.defaultTimeout = 100;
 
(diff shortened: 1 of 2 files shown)


zhijinli:
LGTM

request.js
```diff
@@ -80,6 +80,23 @@ inherits(TChannelRequest, EventEmitter);
...
+    this.timeout = timeout;
+    this.peer = peer;
+    this.waitForIdentifiedSlot = waitForIdentifiedSlot;
+}
+
+RequestOperation.prototype.onTimeout = function onTimeout(now) {
+    if (this.waitForIdentifiedSlot !== null &&
+        this.waitForIdentifiedSlot !== -1
+    ) {
+        this.peer.stopWaitingForIdentified(this.waitForIdentifiedSlot);
```
syyang: does `stopWaitingForIdentified` not do the null or -1 check?

request.js
```diff
@@ -80,6 +80,23 @@ inherits(TChannelRequest, EventEmitter);
...
+function RequestOperation(req, timeout, peer, waitForIdentifiedSlot) {
+    this.req = req;
+    this.timeout = timeout;
+    this.peer = peer;
+    this.waitForIdentifiedSlot = waitForIdentifiedSlot;
+}
+
+RequestOperation.prototype.onTimeout = function onTimeout(now) {
+    if (this.waitForIdentifiedSlot !== null &&
+        this.waitForIdentifiedSlot !== -1
```
syyang: maybe default `waitForIdentifiedSlot` to `-1` above?

request.js
```diff
@@ -304,9 +321,25 @@ TChannelRequest.prototype.resend = function resend() {
...
+        self.elapsed = now - self.start;
+        var timeout = self.timeout - self.elapsed;
+        var reqOp = new RequestOperation(
+            self, timeout, peer, waitForIdentifiedSlot
+        );
+        reqTimeout = self.channel.timeHeap.update(reqOp, now);
+    }
 
     function onIdentified(err) {
+        reqTimeout.cancel();
```
syyang: some comment on why cancel needs to be called will be useful. isn't `reqTimeout` undefined at this point?

request.js
Raynos: Sure.

syyang:
lgtm. thanks!

PR #345, opened 2016-11-07
peer: handle edge case where identified descriptor is null

Raynos:
Because of edge cases there is a race between connection
identification and timeouts.

A timeout may still fire whilst a connection has already
been estabilished.

We need a null check here and do a no-op.

r: @zhijinli @proflayton


proflayton:
lgtm

zhijinli:
LGTM

Raynos:
Cut v3.9.7

PR #346, opened 2016-11-15
Updated peer score strategy to not fully connect

Raynos:
This updates the peer selection score strategy to mark
unconnected peers as the lowest score.

This means that we will not fully connect and will
in general use 1 peer most of the time.

The only time more then 1 peer is used per sub channel:

 - minConnections is configured for p2p
 - hyperbahn is used and extra inbound peers are estabilished.

r: @kriskowal @prashantv @jialiehu

### DIFF
--- a/peer_score_strategies.js
+++ b/peer_score_strategies.js
@@ -60,12 +60,12 @@ PreferOutgoing.prototype.getScoreRange = function getScoreRange() {
     switch (tier) {
         default:
             /* falls through */
+        case PreferOutgoing.UNCONNECTED:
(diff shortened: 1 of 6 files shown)


albertyfwu:
not familiar with tchannel code, but thanks for doing this

peer_score_strategies.js
```diff
@@ -60,12 +60,12 @@ PreferOutgoing.prototype.getScoreRange = function getScoreRange() {
     switch (tier) {
```
kriskowal: please update comment describing ranges^

PR #347, opened 2016-11-22
Make the test suite compat with node4...

Raynos:
Had to fix one issue in the `socket.buffer` API to access the underlying write queue.

Other then that its lots of test massaging to make travis happy. Travis is really annoying.

r: @rf @kriskowal


test/peer-file-watcher.js
```diff
@@ -331,7 +331,7 @@ function sendRequests(channel, count, cb) {
                     as: 'raw'
                 }
             }).send('echo', 'a', 'body', onResponse);
-        });
+        }, Math.random() * 1000);
```
rf: Interesting choice here, why do we need to wait a random amount of time?

Raynos: With batched setImmediate in node4 the test fires all the outbound requests in next tick instead of in the next N ticks.

I want to evenly spread out the socket allocations over N ticks of time.

connection.js
```diff
@@ -554,19 +554,29 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
     }
 };
 
+function getSocketBuffer(socket) {
```
kriskowal: document the reason for this function and note which version each branch caters to.

package.json
```diff
@@ -29,7 +31,7 @@
   },
   "dependencies": {
     "bufrw": "^1.2.1",
-    "crc": "^3.2.1",
+    "crc": "3.2.1",
```
kriskowal: might be good to leave a note somewhere for why we’re pinning crc

test/argstream.js
```diff
@@ -226,25 +227,53 @@ function writeFrames(frames, s, callback) {
     function eachFrame(i) {
         var stream = s['arg' + writingArg];
         if (i >= frames.length) {
-            if (stream) stream.end();
+            if (stream) {
+                console.log('end(' + writingArg + ')');
```
kriskowal: remove console.log

test/peer-file-watcher.js
kriskowal: document intent

package.json
Raynos: It broke on master, now I'm pinning it because upstream broke compat in patch version.

PR #350, opened 2017-03-21
Document request options

kriskowal:
This change has become broader and now adjusts the information architecture for discoverability. I’ve improved coverage over sub-channels with sending and handling requests.


Raynos:
lgtm.

PR #353, opened 2017-11-15
v2/handler: handle case where req is null

Raynos:


hguanyao:
lgtm

Raynos:
Tests pass locally; im gonna ignore jenkins.

Issue #355, opened 2018-05-02
Node 10 Support


lxe:
tchannel-node fails to build on Node 10 Stable due to farmhash 1.1.0 dependency

Raynos:
open an issue on farmhash.

Raynos:
farmhash @ latest supports node10 but it probably does not support node0.10.32 ; we may need conditional dependencies based on node versions.

ajbogh:
#356 should fix this issue.

ajbogh:
Node 0.10.0 works as long as a couple of flags are defined with the right values. The error with farmhash stems from the Mac version, and a lack of the `utility` class for [farmhash's header file](https://github.com/lovell/farmhash/blob/v1.1.0/src/upstream/farmhash.h).

Use flags with npm rebuild (replace `rebuild` with `install` if desired):

```
CXXFLAGS="-mmacosx-version-min=10.9" LDFLAGS="-mmacosx-version-min=10.9" npm rebuild
```

To be clear, this fixes Node 0.10.0, which is not the same as Node 10. Node 10's fix is in #356.

PR #356, opened 2019-04-03
Updated dependencies to be compatible with Node 9+.

ajbogh:


ajbogh:
Tested `npm install` as well as `npm rebuild` with Node 10.15.3.

kriskowal:
Have you also tested with Node.js 0.10?

ajbogh:
@kriskowal 0.10.0 still produces the farmhash error reported in #355. There are no new errors beyond that, or at least the existing exception prevents new errors from displaying.

ajbogh:
@kriskowal, I tested with Node 0.10.0 and it works as long as a couple of flags are defined with the right values. The error from #355 stems from the Mac version, and a lack of the `utility` class for [farmhash's header file](https://github.com/lovell/farmhash/blob/v1.1.0/src/upstream/farmhash.h).

Using flags with npm rebuild (replace `rebuild` with `install` if desired):

```
CXXFLAGS="-mmacosx-version-min=10.9" LDFLAGS="-mmacosx-version-min=10.9" npm rebuild
```

Farmhash has updated their code to v2.1.0 and set the minimum Node version to >4.5.0, so the error from #355 wouldn't be valid with newer dependency versions. Perhaps it would be better to upgrade dependencies rather than fixing bugs for extremely outdated and non-supported versions of Node.

PR #360, opened 2020-01-16
Withdraw support for legacy Node.js versions

kriskowal:
The postinstall hook for binary dependencies was a good idea but interfered with web platform build systems. This change prepares for a major version bump by replacing the postinstall hook with concrete dependencies on the recent versions and reducing the scope of continuous integration. On a separate branch, we will do the same for a patch release to preserve support for legacy versions.

This also presents a rare opportunity to upgrade dependencies without risking breaking changes. This brings all dependencies except eslint to their current best version. Eslint is too hard. Istanbul deprecated its CLI in favor of NYC.


.travis.yml
```diff
@@ -4,21 +4,13 @@ jobs:
   include:
     - node_js: "12"
       env: CXX=g++-4.8 TCHANNEL_TEST_CONFIG= NPM_TEST_SCRIPT=travis
-    - node_js: "5"
```
Raynos: 10 is still LTS we can test on 10 and 12. I think 8 left maintenance so it’s not needed

as/json.js
```diff
@@ -70,7 +69,7 @@ function send(endpoint, head, body, callback) {
     var outreq = self.channel.request(self.reqOptions);
     var headers = head;
     if (self.reqOptions.defaultRequestHeaders) {
-        headers = extend(self.reqOptions.defaultRequestHeaders, head);
+        headers = {...self.reqOptions.defaultRequestHeaders, ...head};
```
Raynos: Nice

PR #361, opened 2020-01-17
Expressly depend on binary dependencies

kriskowal:
This change removes the postinstall hook and replaces it with express dependencies on farmhash and sse4-crc32. This change targets the v3 release train and will not land on master.


Raynos:
Seems legit.

PR #362, opened 2020-04-21
Enable timeHeap configuration.

suraksh:
Currently sub-channel depends on the global time-heap and this will increase the min-heap array size which is used to track the timers. In order to reduce the min heap array size, sub-channel can pass down the timeHeap from the opts.


Raynos:
@srivastavankit How's the Java going >:D

PR #366, opened 2021-01-24
Remove exclusion of 0.0.0.0 as invalid ip address

Choongkyu:
The exclusion of 0.0.0.0 blocks the ability to refer to localhost ip the way it is done on nodejs's idiom of not requiring the server's hostname to be explicitly defined.

### DIFF
--- a/host-port.js
+++ b/host-port.js
@@ -26,17 +26,13 @@ module.exports.validateHostPort = validateHostPort;
 module.exports.validateHost = validateHost;
 module.exports.validatePort = validatePort;
 
-function validateHost(host, allowEmphemeral) {
+function validateHost(host) {
     if (typeof host !== 'string') {
         return 'Expected host to be a string';
     }
 
     if (!validIPv4.test(host)) {
         return 'Expected host to contain IPv4';
-    }
-
-    if (!allowEmphemeral && host === '0.0.0.0') {
-        return 'Expected host to not be 0.0.0.0';
     }
 
     return null;


Raynos:
Tchannel cannot bind on localhost. 

The protocol requires exchanging host port and listening on a public IP

Choongkyu:
> Tchannel cannot bind on localhost.
> 
> The protocol requires exchanging host port and listening on a public IP

For context, the changes in this PR enable me to use this code on the server side:
```
  server.listen(parseInt(process.env.TCHANNEL_PORT), "0.0.0.0", function onListen() {
    console.log("tchannel server listening on 9090");
  });
```
and on the client side calling the server:
```
outbound := tchannelTransport.NewSingleOutbound("auth-svc:9090") // resolves to kubernetes cluster ip 10.103.220.182
dispatcher := yarpc.NewDispatcher(yarpc.Config{
	Name: "tickets-client",
	Outbounds: yarpc.Outbounds{
		"auth": {
			Unary: outbound,
		},
	},
})
if err := dispatcher.Start(); err != nil {
	fmt.Println("dispatcher failed to start")
	panic(err)
}
rootCtx := context.Background()
ctx, cancel := context.WithTimeout(rootCtx, time.Second)
defer cancel()
client := authclient.New(dispatcher.ClientConfig("auth"))
userResponse, err := client.CurrentUser(ctx)
```
I'm not exactly sure what the remarks are addressing, but I think that if there is some erroneous behavior that emerges from passing `0.0.0.0` as the listening ip, this issue can be caught and bubble up downstream anyway.

I'd be open to alternative means of addressing the issue I have in lieu of this PR, but removing the host validation seems like the simplest way of unblocking me.  Also, I noticed that the host validation includes an argument `allowEmphemeral`, but I wasn't seeing any instance of it being used.  I'd be open to any recommended changes for using this so the user is able to decide on whether to enable or disable the validation for `0.0.0.0`.

Raynos:
> when a server is told to listen on 0.0.0.0 that means "listen on every available network interface"

You want to listen on all devices, you can use a module like `my-local-ip` to listen on the a public device.

It basically reads `require('os').networkInterfaces()` and finds a useful network interface to listen to. ( https://github.com/uber-archive/hyperbahn/blob/3b5ebee4fdf815a8f95d8e84bbaedf3fa92d26a9/lib/local_ip.js#L25 ).

Listening on `0.0.0.0` for a tcp server in nodejs is fine, but tchannel requires a known IP to use in the connection handshake for the wire protocol so `0.0.0.0` is blocked.

The simplest change in your nodeJS application is to find the external IP and listen on that. Alternatively, if you want to listen to `127.0.0.1` then you can do that too.

Choongkyu:
ah, gotcha. thank you! I'll make the recommended and close this out.

